PR #3, opened 2020-03-19
Add back missing LIBRARY entry in package bind.

jbyibm:


jbyibm:
@dangeville-n @dennis-behm @drbruce-git for review.

PR #9, opened 2020-04-01
Set the default gateway log level.

jbyibm:


dennis-behm:
Hi Jean-Yves,
shouldn't we use the default, which is defined from zOS?
This is 0, according to this KC page:
https://www.ibm.com/support/knowledgecenter/SSLTBW_2.4.0/com.ibm.zos.v2r4.f54pc00/isppccea5.htm
Dennis

dennis-behm:
@jbyibm Can you quickly elaborate, what you have seen as a behaviour, when the default gateway log level is not provided?

Issue #16, opened 2020-05-06
Move build data sets creation to language scripts


drbruce-git:
Currently all build data sets required by zAppBuild language scripts are created in the build.groovy initialize method even if the build does not actually build all the languages.  While this is not a major issue for team builds, it is mostly unnecessary for user builds and for large customer applications who have introduced more language scripts, could cause problems with dozens of unused data sets created for each developer.  Moving the build data set creation to the language scripts should eliminate unused data set creation . . .

**however . . .**

we will also need to change the build script iteration logic to skip languages that have no files in the sub-list.  Currently zAppBuild runs all language scripts even if the build sub-list is empty.

drbruce-git:
Resolved via https://github.com/IBM/dbb-zappbuild/pull/20

PR #20, opened 2020-05-27
zAppBuild Issue 12 - Support Absolute Path

wmathis:


build.groovy
```diff
@@ -107,11 +114,7 @@ def initializeBuildProcess(String[] args) {
 	// verify/create build data sets required by each language script
 	if (props.languagePropertyQualifiers) {
 		props.languagePropertyQualifiers.trim().split(',').each { lang ->
-			if (props."${lang}_srcDatasets")
-				createDatasets(props."${lang}_srcDatasets".split(','), props."${lang}_srcOptions")
-		
-			if (props."${lang}_loadDatasets")
-				createDatasets(props."${lang}_loadDatasets".split(','), props."${lang}_loadOptions")
+		        buildUtils.getLanguagePropertyQualifiers(lang)
```
dennis-behm, line 117: While this is invoked during the initializationPhase of the build, it will call the getLanguagePropertyQualifiers for all languages defined. However the method getLanguagePropertyQualifiers will also create the datasets.

This means, when performing a UserBuild on a cobol file, it will generate all datasets for all other languages as well. This is what we wanted to avoid per #16

drbruce-git:
@drice-us FYI

Issue #23, opened 2020-06-23
ImpactBuild fixGitDiffPath not working with some configuration 


jbyibm:
The new method fixGitDiffPath  is not working with some configuration. It only works if the `applicationSrcDirs=${application}`. But if we want to specify folders under the `${application}` to only scan source folders that we want to build (to avoid scanning the whole  `${application}` folder it does not work. In this case impact build find nothing. Example config:
>applicationSrcDirs=nazare-demo-genapp/base/src/cobol,nazare-demo-genapp/base/src/bms

PR #24, opened 2020-06-23
ImpactBuild fixGitDiffPath not working with some configuration

jbyibm:
The new method fixGitDiffPath  is not working with some configuration. It only works if the `applicationSrcDirs=${application}`. But if we want to specify folders under the `${application}` to only scan source folders that we want to build (to avoid scanning the whole  `${application}` folder it does not work. In this case impact build find nothing. Example config:
>applicationSrcDirs=nazare-demo-genapp/base/src/cobol,nazare-demo-genapp/base/src/bms


jbyibm:
See: https://github.com/IBM/dbb-zappbuild/issues/23
@dennis-behm @drbruce-git for review.

Issue #30, opened 2020-08-05
Add BZUSAMP library to necessary dataset definitions


dennis-behm:
COBOL.groovy adds the BZUSAMP library, and the template for dataset definitions misses this entry.

https://github.com/IBM/dbb-zappbuild/blob/35adba0136c4f5c9730b847d3e59d5dfea9107a7/build-conf/datasets.properties#L58

dennis-behm:
Delivered

PR #34, opened 2020-08-13
Improve zunit documentation and property management

dennis-behm:
- introducing required build properties for validation
- improved documentation in readme.mds
- minor fixes (missing dataset def) #30

### DIFF
--- a/build-conf/datasets.properties
+++ b/build-conf/datasets.properties
@@ -56,3 +56,6 @@ SEQAMOD=
 
 # Optional IDz Load Library. Example: FEL.V14R0M0.SFELLOAD
 SFELLOAD=
+
+# Optional IDZ zUnit / WAZI VTP library containing necessary copybooks. Example : RDZ.V14R2.SBZUSAMP
+SBZUSAMP=
--- a/build-conf/datasets.properties
+++ b/build-conf/datasets.properties
@@ -57,5 +57,5 @@ SEQAMOD=
 # Optional IDz Load Library. Example: FEL.V14R0M0.SFELLOAD
 SFELLOAD=
 
-# Optional IDZ zUnit / WAZI VTP library containing necessary copybooks. Example : RDZ.V14R2.SBZUSAMP
+# Optional IDZ zUnit / WAZI VTP library containing necessary copybooks. Example : FEL.V14R2.SBZUSAMP
 SBZUSAMP=
--- a/samples/application-conf/README.md
+++ b/samples/application-conf/README.md
@@ -25,7 +25,10 @@ Location of file properties, script mappings and file level property overrides. 
 
 Property | Description 
 --- | --- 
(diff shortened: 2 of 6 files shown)


build-conf/datasets.properties
```diff
@@ -55,4 +55,7 @@ REFERAL=
 SEQAMOD=
 
 # Optional IDz Load Library. Example: FEL.V14R0M0.SFELLOAD
-SFELLOAD=
\ No newline at end of file
+SFELLOAD=
+
+# Optional IDZ zUnit / WAZI VTP library containing necessary copybooks. Example : RDZ.V14R2.SBZUSAMP
```
jbyibm, line 60: Can you put the same sample dataset as the README.md

samples/application-conf/README.md
```diff
@@ -25,7 +25,10 @@ Location of file properties, script mappings and file level property overrides.
 
 Property | Description 
 --- | --- 
-dbb.scriptMapping | DBB configuration file properties associtation build files to language scripts
+dbb.scriptMapping | DBB configuration file properties association build files to language scripts
+dbb.scannerMapping | DBB scanner mapping to overwrite the file scanner. File property
+cobol_testcase | File property to indicate a generated zUnit cobol test case to use a different set of source and output libraries
```
jbyibm, line 30: This is not clear. But maybe because I have minimal knowledge in zUnit for now.

samples/application-conf/file.properties
```diff
@@ -15,3 +15,7 @@ dbb.scriptMapping = ZunitConfig.groovy :: **/*.bzucfg
 #
 # Scanner mappings for application programs that require a custom scanner
 dbb.scannerMapping = ZUnitConfigScanner :: **/*.bzucfg
+
+#
+# file mapping for generated zUnit Test case programs (Cobol) to use a seperate set of libraries
+# cobol_testcase = true :: **/testcase/*.cbl
```
jbyibm, line 21: In case the value `cobol_testcase ` is not defined in `file.properties` what about the following code in `Cobol.groovy` ?
```
	// Check if this a testcase
	isZUnitTestCase = (props.getFileProperty('cobol_testcase', buildFile).equals('true')) ? true : false
```

if `props.getFileProperty('cobol_testcase', buildFile)` return null you will have an NullPointerException exception.

jbyibm (changes requested):

samples/application-conf/file.properties
dennis-behm: Apparently, I don't receive a NPE in the case of not providing a cobol_testcase file property.

build-conf/datasets.properties
dennis-behm: See recent commit

jbyibm (approved):

PR #37, opened 2020-09-01
Fix zUnit Test PDS creation/Fix zUnit missing report in case of failure

jbyibm:
@dennis-behm @ricedavida 
This is minor fixes that you may already work on this.

NOTES:
- Before calling the JCL we may need to remove the report in the PDS member to be sure to not have a report from a previous build (not done).

### DIFF
--- a/build-conf/Cobol.properties
+++ b/build-conf/Cobol.properties
@@ -45,6 +45,8 @@ cobol_loadOptions=cyl space(1,1) dsorg(PO) recfm(U) blksize(32760) dsntype(libra
 cobol_tempOptions=cyl space(5,5) unit(vio) blksize(80) lrecl(80) recfm(f,b) new
 cobol_printTempOptions=cyl space(5,5) unit(vio) blksize(133) lrecl(133) recfm(f,b) new
 
+#
+# List the data sets for tests that need to be created and their creation options
 cobol_test_srcDatasets=${cobol_testcase_srcPDS}
 cobol_test_srcOptions=cyl space(1,1) lrecl(80) dsorg(PO) recfm(F,B) dsntype(library)
 
--- a/build-conf/Cobol.properties
+++ b/build-conf/Cobol.properties
@@ -45,11 +45,11 @@ cobol_loadOptions=cyl space(1,1) dsorg(PO) recfm(U) blksize(32760) dsntype(libra
 cobol_tempOptions=cyl space(5,5) unit(vio) blksize(80) lrecl(80) recfm(f,b) new
 cobol_printTempOptions=cyl space(5,5) unit(vio) blksize(133) lrecl(133) recfm(f,b) new
 
-cobol_testsrcDatasets=${cobol_testcase_srcPDS}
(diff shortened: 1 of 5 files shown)


languages/Cobol.groovy
```diff
@@ -20,6 +20,11 @@ buildUtils.assertBuildProperties(props.cobol_requiredBuildProperties)
 def langQualifier = "cobol"
 buildUtils.createLanguageDatasets(langQualifier)
 
+if (props.runzTests == "True") {
```
dennis-behm, line 23: It appears, that the property is now tested twice, it is already part of `createLanguageDatasets` , where the language cobol_testDatasets is used.

https://github.com/IBM/dbb-zappbuild/blob/5b283e9a8a13558afa2f8924448e1727b9b525c0/utilities/BuildUtilities.groovy#L357
So, there is definitely a defect in the `createLanguageDatasets` method.

However, with the proposed changes, this section would no longer be required. So, I think we should remove it from the `createLanguageDatasets` method

build-conf/Cobol.properties
```diff
@@ -45,11 +45,11 @@ cobol_loadOptions=cyl space(1,1) dsorg(PO) recfm(U) blksize(32760) dsntype(libra
 cobol_tempOptions=cyl space(5,5) unit(vio) blksize(80) lrecl(80) recfm(f,b) new
 cobol_printTempOptions=cyl space(5,5) unit(vio) blksize(133) lrecl(133) recfm(f,b) new
 
-cobol_testsrcDatasets=${cobol_testcase_srcPDS}
-cobol_testsrcOptions=cyl space(1,1) lrecl(80) dsorg(PO) recfm(F,B) dsntype(library)
+cobol_test_srcDatasets=${cobol_testcase_srcPDS}
```
dennis-behm, line 48: we could need to update the readme.md for the build config as well https://github.com/IBM/dbb-zappbuild/tree/development/build-conf

languages/ZunitConfig.groovy
```diff
@@ -95,7 +95,12 @@ buildUtils.createLanguageDatasets(langQualifier)
 	zUnitRunJCL.saveOutput(logFile, props.logEncoding)
 
 	// Store Report in Workspace
-	new CopyToHFS().dataset(props.zunit_bzureportPDS).member(member).file(reportLogFile).hfsEncoding(props.logEncoding).append(false).copy()
+	try {
```
dennis-behm, line 98: I think, we should move this into the section where the jobRC is evaluated. Only with a valid RC, we should try to store it. Otherwise we might retrieve a previously generated report.

dennis-behm (changes requested):
@jbyibm can you please have a look to the comments?

languages/Cobol.groovy
dennis-behm: I have removed the section in BuildUtilities.groovy

dennis-behm (approved):
@drbruce-git Please have a final review of the changes.

drbruce-git (approved):

Issue #43, opened 2020-09-28
Cobol.groovy fails during user build due to non-existent BMS copybook dataset


drbruce-git:
Build dataset creation was recently moved from `build.groovy` to the individual language scripts to avoid having user build creating dozens of unused datasets.  However the `Cobol.groovy` build script includes the BMS copybook dataset in its SYSLIB concatenation.  Its conditional subject to the existence of BMS copybook name property but its defined in the `build-conf` folder, so it always exists even if the dataset has not been created.  We need to use another condition.  Perhaps checking the build order property for the BMS entry.  Or just using JZOS to test if the dataset exists.

PR #44, opened 2020-09-30
Test if BMS datasets exit

dennis-behm:
While the dataset creation was moved to the language scripts, we might see builds failing due to a missing dataset - e.q. the application does not make use of BMS. See #43


drbruce-git (changes requested):
@dennis-behm  I tested the condition as written and it does not work as it is missing single quotations.  Also you should verify that the props.bms_cpyPDS property exists a value before executing the dsExists.  Please replace with the following condition.  I've tested it and it works:
```
if (props.bms_cpyPDS && ZFile.dsExists("'${props.bms_cpyPDS}'"))
```

drbruce-git (approved):

PR #46, opened 2020-10-09
zunit cleanup

dennis-behm:
- uses the standard boolean representation
    a) single cmdline option, b) application-conf uses `true` rather than `True`
- fixes the classification return codes reported from the zunit runner
@jbyibm @drbruce-git

### DIFF
--- a/build.groovy
+++ b/build.groovy
@@ -39,7 +39,7 @@ else {
 		println("** Invoking build scripts according to build order: ${props.buildOrder}")
 		String[] buildOrderList = props.buildOrder.split(',')
 		String[] testOrderList;
-		if (props.runzTests == "True") { 
+		if (props.runzTests) { 
 			println("** Invoking test scripts according to test order: ${props.testOrder}")
 			testOrderList = props.testOrder.split(',')
 		}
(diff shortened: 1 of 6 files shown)


build.groovy
```diff
@@ -39,7 +39,7 @@ else {
 		println("** Invoking build scripts according to build order: ${props.buildOrder}")
 		String[] buildOrderList = props.buildOrder.split(',')
 		String[] testOrderList;
-		if (props.runzTests == "True") { 
+		if (props.runzTests) {
```
jbyibm, line 42: It is enough ? Or should we have this ?
if (props.runzTests && props.runzTests .toBoolean())

utilities/BuildUtilities.groovy
```diff
@@ -331,7 +331,7 @@ def relativizeFolderPath(String folder, String path) {
  * getScannerInstantiates - returns the mapped scanner or default scanner
  */
 def getScanner(String buildFile){
-	if (props.runzTests == "True") {
+	if (props.runzTests) {
```
jbyibm, line 334: It is enough ? Or should we have this ?
if (props.runzTests && props.runzTests .toBoolean())

languages/Cobol.groovy
```diff
@@ -17,10 +17,11 @@ println("** Building files mapped to ${this.class.getName()}.groovy script")
 // verify required build properties
 buildUtils.assertBuildProperties(props.cobol_requiredBuildProperties)
 
+// create language datasets
 def langQualifier = "cobol"
 buildUtils.createLanguageDatasets(langQualifier)
 
-if (props.runzTests == "True") {
+if (props.runzTests) {
```
jbyibm, line 24: It is enough ? Or should we have this ?
if (props.runzTests && props.runzTests .toBoolean())

jbyibm (changes requested):

build.groovy
dennis-behm: Agreed, we need to test the boolean value of it!

dennis-behm:
@jbyibm requested changes delivered

jbyibm (approved):

jbyibm:
@dennis-behm It is possible to merge this PR ?

PR #48, opened 2020-11-12
Expose list of deleted files

dennis-behm:
zAppBuild already calculates the deleted files to cleanup collections. 

This PR contributes that a list of deleted files is written as `deletedFilesList` to the build workspace, if deletions were found during impactCalculation. This is available for `impactBuilds`.


jbyibm (approved):
Look good.

Issue #49, opened 2020-11-17
Simple way to add existing libraries to syslib concatenation


dennis-behm:
Current state: If you would like to add a set of libraries to the syslib concatenation for the compiler or the linker, it requires to modify the language script and probably introduce a new dataset property.

We would like to make it simpler to add existing libraries to the concatenation in a very generic way. Possible use cases are such as:
- adding a new Copy Library or Link library for 3rd party software.
- during the transition period in a phased migration to git, to add the libraries of the legacy system and make copybooks available to the dbb build. 
- possibility to pick up copybooks from other application components managed in git, which have published their copybooks into a shared PDS library.

dennis-behm:
Delivered as part of #50

PR #50, opened 2020-11-17
Simple way to add existing libraries to syslib concatenation

dennis-behm:
This is related to #49 _Simple way to add existing libraries to syslib concatenation_ 

By providing a comma-separated list of existing PDS libraries a user is able to contribute additional existing libraries to the syslib concatenation for ASM, Cobol, PLI or LinkEdit. DD statements for these libraries are added automatically by the language script -between the application libraries populated by zappbuild  and the system libraries for the middleware components.


jbyibm:
It is possible to reduce the number of commits 😄. We have aaaa, bbbbb, Changed *.groovy.

Issue #51, opened 2020-11-20
When we do the linkedit during the cobol compilation links edit datasets are not created


jbyibm:
Example: for an IMS application the  linkedit step failed with this error:

```
*** Building file nazare-demo-imsapp/COBOL/LOADHIST.cbl
IKJ56228I DATA SET NAZARE.WAZI.IMSAPP.LINK NOT IN CATALOG OR CATALOG CAN NOT BE ACCESSED    
org.codehaus.groovy.runtime.InvokerInvocationException: com.ibm.dbb.build.BuildException: java.io.IOException: com.ibm.dbb.build.BuildException: BGZTK0016E An error occurred running BPXWDYN command 'free dd(SYS00102)'.
```

PR to come.

PR #52, opened 2020-11-20
When we do the linkedit during the cobol compilation links edit datasets are not created

jbyibm:
https://github.com/IBM/dbb-zappbuild/issues/51

@dennis-behm FYI


dennis-behm:
Thanks @jbyibm

PR #56, opened 2020-11-26
Creation of "$hlq.TEST.COBOL" dataset conditional to "isZUnitTestCase".

EmanuP:
When using IDz to build cobol testcase the option runzTests is not toggled, because we are just compiling the program and not executing the test. Therefore I modified the script to create/verify the dataset each time we ask DBB to build a testcase.


jbyibm (approved):
Looks good.

dennis-behm:
The current code will run through the dataset creation / verification method for **each** file in the buildList which is mapped to an cobol_testcase .

Probably we move it to the same area where the datasets are verified / created? At least, we should make sure, that it runs only one time?

PR #58, opened 2020-12-04
Add code coverage collector in the zUnit JCL

jbyibm:
@dennis-behm FYI


dennis-behm (approved):
Thank you for adding code coverage support for zUnit!

Issue #59, opened 2020-12-08
Ability to skip impact analysis for a set of files


dennis-behm:
In some cases, users would like to skip the impact analysis phase for a set of copybooks, include or macros files. This could be because this file is used in basically all the programs, and changes are assumed to be backward compatible.

dennis-behm:
merged to development via #60

PR #60, opened 2020-12-08
Ability to skip impact analysis for a set of files

dennis-behm:
Use case is described in #59 .

### DIFF
--- a/utilities/ImpactUtilities.groovy
+++ b/utilities/ImpactUtilities.groovy
@@ -59,18 +59,16 @@ def createImpactBuildList(RepositoryClient repositoryClient) {
 			if (props.verbose) println "** Found build script mapping for $changedFile. Adding to build list"
 		}
 
-		// perform impact analysis on changed file
-		if (props.verbose) println "** Performing impact analysis on changed file $changedFile"
-		ImpactResolver impactResolver = createImpactResolver(changedFile, props.impactResolutionRules, repositoryClient)
+		// check if impact calculation should be performed, default true
+		if (shouldCalculateImpacts(changedFile)){
 
-		// get excludeListe
-		List<PathMatcher> excludeMatchers = createExcludePatterns()
-		
-		// get
-		List<PathMatcher> nonImpactingFiles = createPathMatcherPattern(props.skipImpactCalcuationList)
+			// perform impact analysis on changed file
+			if (props.verbose) println "** Performing impact analysis on changed file $changedFile"
+			ImpactResolver impactResolver = createImpactResolver(changedFile, props.impactResolutionRules, repositoryClient)
 
-		// manage exceptions to bypass impact calculation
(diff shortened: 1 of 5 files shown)


utilities/ImpactUtilities.groovy
```diff
@@ -386,6 +392,56 @@ def matches(String file, List<PathMatcher> pathMatchers) {
...
+	//	{
+	//		Properties impactProperties = new Properties()
+	//		propertyFile.withInputStream { impactProperties.load(it) }
+	//		impactChanges = impactProperties.get('impact_changes')
+	//		if (impactChanges != null){
+	//			if (impactChanges.toBoolean()==false) return false
+	//		}
+	//	}
+	//	return true // default
+}
```
jbyibm, line 425: Can you remove this code in comments ?

utilities/ImpactUtilities.groovy
```diff
@@ -386,6 +392,56 @@ def matches(String file, List<PathMatcher> pathMatchers) {
...
+	if (property) {
+		property.split(',').each{ filePattern ->
+			if (!filePattern.startsWith('glob:') || !filePattern.startsWith('regex:'))
+				filePattern = "glob:$filePattern"
+			PathMatcher matcher = FileSystems.getDefault().getPathMatcher(filePattern)
+			pathMatchers.add(matcher)
+		}
+	}
+	return pathMatchers
+}
```
jbyibm, line 443: I think that this method can be merged with createExcludePatterns the 2 are very similar. Maybe createPathMatcherPattern can be use in both cases with the property parameter.

jbyibm (changes requested):
See comments.

jbyibm (approved):
Look good

utilities/ImpactUtilities.groovy
dennis-behm: Makes sense, I have updated that.

utilities/ImpactUtilities.groovy
dennis-behm: The initial idea was to have a choice, but you are right. We should move this sample snippet to https://github.com/ibm/dbb

dennis-behm:
### Sample usage

Specify the files, for which you would skip the impact analysis at 
https://github.com/IBM/dbb-zappbuild/blob/b55da1e5ea756e805f4581048f5105942c1c2708/samples/MortgageApplication/application-conf/application.properties#L27-L31

**Sample configuration** to skip analysis for the impacts for epsmtout.cpy. Please be aware, that this uses the glob syntax to identify them:
```
skipImpactCalculationList=**/epsmtout.cpy
```

During an **impactBuild** the impact analysis is skipped, which is reported in the build log in verbose mode:
```
..
** Impact analysis for MortgageApplication/copybook/epsmtout.cpy has been skipped due to configuration.
..
```

PR #65, opened 2020-12-16
Standard XML parser of Groovy is used to get playback file name.

EmanuP:
I have modified ZunitConfig.groovy to use the standard XML parser instead of the split method to get the playback file name.


dennis-behm (approved):
Thanks for the update!

PR #76, opened 2021-01-13
Fixed typo in property name

franktuma:
Updated assembler_requiredBuildProperties to match what is being used in Assembler.groovy


dennis-behm:
Thank you!

PR #78, opened 2021-01-25
Contains tests

EmanuP:
Cobol.groovy script has been modified to build cobol testcases independently of the flag "runzTests".

The code performance was improved with respect to the previous pull request.

### DIFF
--- a/languages/Cobol.groovy
+++ b/languages/Cobol.groovy
@@ -340,7 +340,8 @@ def getRepositoryClient() {
 	return repositoryClient
 }
 
-def buildListContainsTests(List<String> buildList) {
+boolean buildListContainsTests(List<String> buildList) {
+	boolean testFound = false
 	// iterate through build list
 	buildList.each { buildFile ->
 	
@@ -349,8 +350,8 @@ def buildListContainsTests(List<String> buildList) {
 		
 		// create test datasets
 		if (isZUnitTestCase) {
-			return true
+			testFound = true
 		}
 	}
-	return false
+	return testFound
 }
--- a/languages/Cobol.groovy
+++ b/languages/Cobol.groovy
@@ -343,15 +343,12 @@ def getRepositoryClient() {
 def buildListContainsTests(List<String> buildList) {
 	// iterate through build list
 	buildList.each { buildFile ->
-		println "*** Building file $buildFile"
 	
 		// Check if this a testcase
 		isZUnitTestCase = (props.getFileProperty('cobol_testcase', buildFile).equals('true')) ? true : false
 		
 		// create test datasets
 		if (isZUnitTestCase) {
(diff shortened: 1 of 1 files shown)


languages/Cobol.groovy
```diff
@@ -339,3 +339,18 @@ def getRepositoryClient() {
...
+boolean buildListContainsTests(List<String> buildList) {
+	boolean testFound = false
+	// iterate through build list
+	buildList.each { buildFile ->
+	
+		// Check if this is a testcase
+		isZUnitTestCase = (props.getFileProperty('cobol_testcase', buildFile).equals('true')) ? true : false
+		
+		if (isZUnitTestCase) {
+			testFound = true
```
dennis-behm, line 352: We can directly `return true` if the first test is found, instead of integrating through the entire buildList.

dennis-behm (changes requested):
Hi @EmanuP, thanks for the PR and apologies for the very late response. 
I like to idea to just create the datasets in case there is a test involved.
Can you please have a look to the two comments?

languages/Cobol.groovy
EmanuP: Good idea, this will improve performance.

EmanuP:
@dennis-behm Does the last commit reflect your idea?

languages/Cobol.groovy
```diff
@@ -339,3 +339,17 @@ def getRepositoryClient() {
...
+
+boolean buildListContainsTests(List<String> buildList) {
+	// iterate through build list
+	buildList.each { buildFile ->
+	
+		// Check if this is a testcase
+		isZUnitTestCase = (props.getFileProperty('cobol_testcase', buildFile).equals('true')) ? true : false
+		
+		if (isZUnitTestCase) {
+			return true
```
dennis-behm, line 351: What I understood from the docs, you cannot break/return out from an each{closure} loop, that is why your current method always returns `false` and the datasets are not allocated.

Let's just switch to a `return isZUnitTestCase` at the very end.

dennis-behm (changes requested):
Please see the comment. The current code always returns false.

languages/Cobol.groovy
dennis-behm: @EmanuP 

Can you please replace the method by, I already tested the code. This should be fine:

```
// check build list for test cases
def boolean buildListContainsTests(List<String> buildList) {

	boolean containsZUnitTestCase = buildList.find { buildFile -> props.getFileProperty('cobol_testcase', buildFile).equals('true')}
	
	return containsZUnitTestCase ? true : false
}
```

dennis-behm (approved):

dennis-behm:
@jbyibm, this fixes #81

PR #79, opened 2021-01-25
Z Unit Batch IO

EmanuP:
ZunitConfig.groovy has been modified to handle a testcase with any combination of the following:

1. Playback files
2. IO files (not tested yet with VSAM)


dennis-behm:
With the recent enhancements of the zUnit Framework of IDZ, these changes are no longer required.

PR #95, opened 2021-03-09
add buildResultProperties to the build report data

dennis-behm:
In case of a build connecting via the repoClient to the dbb web application, we add the build result properties to a generic DBB Record named `DBB.BuildResultProperties`

See sample :
image

### DIFF
--- a/build.groovy
+++ b/build.groovy
@@ -473,29 +473,15 @@ def finalizeBuildProcess(Map args) {
 
 	// create build report data file
 	def jsonOutputFile = new File("${props.buildOutDir}/BuildReport.json")
-	println "** Writing build report data to ${jsonOutputFile}"
 	def buildReportEncoding = "UTF-8"
 	def buildReport = BuildReportFactory.getBuildReport()
-	buildReport.save(jsonOutputFile, buildReportEncoding)
-
-	// create build report html file
-	def htmlOutputFile = new File("${props.buildOutDir}/BuildReport.html")
-	println "** Writing build report to ${htmlOutputFile}"
-	def htmlTemplate = null  // Use default HTML template.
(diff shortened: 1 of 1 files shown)


build.groovy
```diff
@@ -516,11 +502,43 @@ def finalizeBuildProcess(Map args) {
...
+	// create build report html file
+	def htmlOutputFile = new File("${props.buildOutDir}/BuildReport.html")
+	println "** Writing build report to ${htmlOutputFile}"
+	def htmlTemplate = null  // Use default HTML template.
+	def css = null       // Use default theme.
+	def renderScript = null  // Use default rendering.
+	def transformer = HtmlTransformer.getInstance()
+	transformer.transform(jsonOutputFile, htmlTemplate, css, renderScript, htmlOutputFile, buildReportEncoding)
+	
+	// save json file
```
drbruce-git, line 530: Another minor issue.  Can you create and save the data file before the html file?  I know it does not matter functionally but since its passed into the html traqnsformer, logically it should be created and saved before the transformer is called.

build.groovy
```diff
@@ -473,29 +473,15 @@ def finalizeBuildProcess(Map args) {
```
drbruce-git, line 473: Minor issue. Can you move the build report data file comment and variables down with the html file variables and where you actually create and save the file?

drbruce-git (changes requested):

dennis-behm:
@drbruce-git
Can you please have a look to the updated PR.
Additionally to your requested changes, I 
- added the giturl into the properties
- moved the println of the update of the buildresult to the end and included the buildreporturl

Many thanks.

drbruce-git (approved):

PR #97, opened 2021-03-15
add db2 and cics preprocessor steps to assembler language script

dennis-behm:
zAppBuild currently only processes assembler_batch. Enhancing to add a dedicated Db2 and Assembler preprocessor step.


dennis-behm:
Changed the state for DRAFT, while this code requires the upcoming PTF, which is not yet published. Depends on updates of the assembler scanner.

drbruce-git (approved):

Issue #101, opened 2021-04-23
Allow file level property overwrite for the ImpactResolution rules


dennis-behm:
There might arise some situations, where the impactResolution rules need to be able to overwrite the impactResolution rule.

https://github.com/IBM/dbb-zappbuild/blob/d0b53524c8d12a01fd2efae4b606c14a188ba072/utilities/ImpactUtilities.groovy#L66-L72

PR #102, opened 2021-04-23
support impactResolutionRules as a file property

dennis-behm:
During impactAnalysis, we currently supply the entire list of impactRules. 

This enhancement will allow you to limit the impactResolutionRules through a file property in `file.properties` or `application.properties`, e.q. you can just use the copybook and link rule in impactAnalysis for COBOL files:

```
impactResolutionRules=[${copybookRule},${linkRule}] :: **/copy/*.cpy,**/cobol/*.cbl
```


drbruce-git (approved):

dennis-behm:
See #101

Issue #113, opened 2021-05-27
Easily identify changed files in an impactBuild scenario


dennis-behm:
Provide an easy way for users to understand the changes in a build. Even while pipeline orchestrators do this, zappbuild should document this in the build report.

As of today, the impact build scenario captures the current commit hash.

dennis-behm:
Idea

Store the link to the compare of the git provider, like:

https://github.com/ibm/dbb-zappbuild/compare/ac81e15900eef5eb87ef12397e2ecec29587e929...257c164ef5dbe2c87102000405be886f8db5d506

dennis-behm:
Merged per #114

PR #114, opened 2021-06-14
Capture link for changed files for impactBuild scenario in dbb build results

dennis-behm:
This implements #113 .

In case of an `--impactBuild` scenario, the dbb build report contains a link to the actual changed files. Similar to the calculation of changed files through the `git diff` between the baseline and current hash - at https://github.com/IBM/dbb-zappbuild/blob/0d2fe5899fb55d89a4c977bbf53259710281386a/utilities/GitUtilities.groovy#L180

Sample output in verbose (adjusted so you actually see the compare view in Github)

```
*** Obtaining hash for directory /u/dbehm/test-zapp/dbb-zappbuild/samples/MortgageApplication
** Setting property :githash:MortgageApplication : 0d2fe5899fb55d89a4c977bbf53259710281386a
** Setting property :giturl:MortgageApplication : https://github.com/dbb-zappbuild/dbb-zappbuild.git
** Setting property :gitchangedfiles:MortgageApplication : https://github.com/ibm/dbb-zappbuild/compare/49401c9bd8873d95855ea07bbf9b982aa87ceac4..0d2fe5899fb55d89a4c977bbf53259710281386a
** Writing build report data to /u/dbehm/test-zapp-app-out/mortgageout/build.20210614.012436.024/BuildReport.json
** Writing build report to /u/dbehm/test-zapp-app-out/mortgageout/build.20210614.012436.024/BuildReport.html
```

`retrieveLastSuccessfulBuildResult()` in buildUtils is introduced to centrally obtain the last successful build result.

### DIFF
--- a/build-conf/README.md
+++ b/build-conf/README.md
@@ -39,6 +39,9 @@ buildListFileExt | File extension that indicates the build file is really a buil
 languagePropertyQualifiers | List of language script property qualifiers. Each language script property has a unique qualifier to avoid collision with other language script properties.
 applicationConfRootDir | Alternate root directory for application-conf location.  Allows for the deployment of the application-conf directories to a static location.  Defaults to ${workspace}/${application}
 requiredBuildProperties | Comma separated list of required build properties for zAppBuild/build.groovy. Build and language scripts will validate that *required* build properties have been set before the script runs.  If any are missing or empty, then a validation error will be thrown.
+continueOnScanFailure | Determine the behavior when facing a scanner failure. true (default) to continue scanning. false will terminate the process. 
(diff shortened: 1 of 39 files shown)


drbruce-git (commented):
So then this feature only supports single repository builds?  How about multiple repository builds?  How are Git login in credentials handled?  Or does this just provide a link and then the customer has to use the Web Browser to log in?

dennis-behm:
Right now it only supports single repo builds, this is correct. We can add multiple repo builds later. It just creates a link which the customer can use in the browser, so no need to care about credentials.

While Jenkins or others CI orchestrators also provide links to commits, there might be a mismatch, if the previous build failed. The link represents the actual changed files which got identified by the build framework.

Issue #115, opened 2021-06-16
Provide a detailed documentation on the different build scenario


dennis-behm:
The different build scenarios are currently documented on this page:

https://github.com/IBM/dbb-zappbuild/blob/development/README.md#build-scope

Let's dedicate a page to provide samples, for each scenario.

dennis-behm:
This can be closed, while it was delivered as part of #140

PR #119, opened 2021-06-22
Document zAppBuild invocation samples including console log

dennis-behm:
Update on the documentation of build invocation examples, related to #115.


drbruce-git:
@dennis-behm  I would prefer not to show console logs with --verbose turned on in the documentation.  It makes the output look too cluttered for successful builds.

PR #137, opened 2021-08-24
Fix wrong deployType for BMS Copy

jbyibm:
@dennis-behm  Can you integrate this small change quickly?


dennis-behm:
I am merging this one to development; we will add it back in later.

PR #138, opened 2021-08-25
Modified rule to perform impact analysis on zUnit Tests

EmanuP:
According to the documentation the impact resolver's initial query returns resources (i.e. test configuration files) that potentially are impacted by changes to the program being test (i.e. the test configuration file). By applying the resolution rules it can be determined that a test is impacted by the specific changed program.

Therefore the logical dependency of the zUnit test of type ZUNITINC and library SYSPROG should be resolved to the cobol source and not to the configuration file.


dennis-behm:
I am closing the PR, we might need to reopen if necessary.

PR #140, opened 2021-08-30
New build type - set the baseline hash in impact builds

dennis-behm:
Depending on the selected branching model, build situations occur, that the implementation of `impactBuild` or the `impactBuild for topic branches` may not be sufficient to drive incremental builds.

One case is a release assembly process following the a standard git flow process, when several features are combined into an incremental release candidate package. In this scenario, the first build for this configuration should determine the changed files of all the contributed features and pass them to the impact analysis. 
In this case, the application team would rather define the baseline reference, which is passed into the changed files analysis processing.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -785,16 +785,15 @@ Cobol compiler parms for MortgageApplication/cobol/epscmort.cbl = LIB,CICS,SQL
 
 ### Perform ImpactBuilds by overwriting the baseline reference for the analysis of changed files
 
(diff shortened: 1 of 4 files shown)


dennis-behm:
@drbruce-git Can you please have a look to this PR, it adds a new option for impact builds

drbruce-git (approved):

dennis-behm:
The documentation of the feature can be found in the README at: https://github.com/IBM/dbb-zappbuild/blob/development/BUILD.md#perform-impact-build-by-providing-baseline-reference-for-the-analysis-of-changed-files

PR #141, opened 2021-09-16
Update zAppBuild to support IDE passed dependency information

lburgess07:
Update zAppBuild to support IDE passed dependency information

In order to increase performance of User Build running on ZD&T, the IDEs can optionally pass dependency information about the program being built to zAppBuild allowing it to skip running dependency resolution which depending on the size and number of build dependencies the program references can be time consuming on ZD&T platforms.
**Requires IDz version 15.0.3**
**Requires IBM Z Open Editor version 1.4.0**

Note: Backwards (user build) compatibility of IDz and Z Open Editor is maintained. zAppBuild will function the same as previously if no --dependencyFile option is passed. 

Changes Made:
- Added -df (--dependencyFile) option to build.groovy which accepts 1 argument, the Absolute or relative (from workspace) path to user build JSON file containing dependency information
- Updated BuildUtilities.copySourceFiles() to, if a user build dependency file is provided, skip dependency resolution and scan by reading dependency and program information from the dependency file uploaded by the IDE. 
- If User Build Dep file is provided, the JSON file is validated for required fields: BuildUtilities.validateDependencyFile()
- If no dependency file is provided (and thus no -df option is passed to build.groovy), then zApp will revert to traditional scan and dependency resolution just as it did previously.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -44,6 +44,10 @@ $DBB_HOME/bin/groovyz build.groovy --workspace /u/build/repos --application app1
 ```
 $DBB_HOME/bin/groovyz build.groovy --workspace /u/build/repos --application app1 --outDir /u/build/out --hlq BUILD.APP1 --fullBuild --cc --cch localhost --ccp 8009 --cco "e=CCPDF"
 ```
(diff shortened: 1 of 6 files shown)


dennis-behm (commented):
@lburgess07 Thanks for the updates!

Issue #149, opened 2021-11-10
Validation of the dbb toolkit version in use


dennis-behm:
The new DBB Toolkit 1.1.2 has recently added some new APIs - e.q. new APIs which will allow the search paths to be simpler and supporting wild cards. 

This requires, that zAppBuild validates the version of the dbb toolkit which is available on the build machine.

Let's start by 
* print out the current toolkit version in use
* prepare for a check of the toolkit, when the build is initialised to match a minimum required dbb toolkit version

PR #150, opened 2021-11-11
Printout and validation of the DBB Toolkit version in use

dennis-behm:
* Provides the printout in `--verbose` tracing of the DBB toolkit which is use . 
```
** zAppBuild running on DBB Toolkit Version 1.1.1 03-Jun-2021 19:39:09
```
* Validates the toolkit against the defined requiredDbbToolkit level in build.properties


drbruce-git (approved):

dennis-behm:
See #149

Issue #151, opened 2021-11-17
New build type to calculate changes flowing back to main branch


dennis-behm:
Depending on the branching strategy, scenarios might occur where you would like to build just these changes on your topic branch which will flow back to the main branch.

The `git diff` cmd allows an interesting diff'ing strategy using a triple-dot diff, which will report the changes flowing back to the main branch.

Some reading material can be found here: [Blog - git-diff-triple-dot-and-double-dot ](https://kimsereylam.com/git/2020/12/11/git-diff-triple-dot-and-double-dot.html)

I propose a to introduce a new build type `mergeBuild` which calculates and builds the changes flowing back to the `mainBuildBranch`.

PR #152, opened 2021-11-17
Build Type mergeBuild

dennis-behm:
This enhancement implements a new build strategy in zAppBuild for topic branches and addresses #151 

The purpose of `--mergeBuild` is to calculate changes which will be merged from the current configuration into the target configuration. 

It uses a `triple-dot git diff` to compare the HEAD to the target remote configuration using the `mainBuildBranch`.

The changes include
* Implementation and Documentation of the new build strategy 
* Test case for the zAppBuild Test Framework

As mentioned in the documentation `--mergeBuild` only builds the changes. **It does not perform impact analysis and is not incremental**; it always processes the list returned by the git diff. Ex.: When you change one file in your topic branch, mergeBuild will process this file. When you change and commit a second file, then both files will be built.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -970,6 +970,7 @@ groovyz dbb-zappbuild/build.groovy --workspace /var/dbb/dbb-zappbuild/samples --
 ```
 <details>
   <summary>Build log</summary>
+
 ```
 + /usr/lpp/dbb/v1r0/bin/groovyz /var/dbb/dbb-zappbuild/build.groovy --sourceDir /var/dbb/dbb-zappbuild/samples --workDir /var/dbb/out/MortgageApplication --hlq DBB.ZAPP.MERGE.BUILD --application MortgageApplication --verbose --mergeBuild --propFiles /var/dbb/dbb-zappbuild-config/build.properties,/var/dbb/dbb-zappbuild-config/datasets.properties
 
@@ -1071,6 +1072,8 @@ Cobol compiler parms for MortgageApplication/cobol/epsmlist.cbl = LIB,CICS
 ** Total files processed : 1
 ** Total build time  : 5.468 seconds
 ```
+
+
 </details>
 
 ### Perform a Scan Source build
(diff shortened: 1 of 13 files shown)


BUILD.md
```diff
@@ -123,6 +129,7 @@ utility options
 - [Perform Impact Build](#perform-impact-build)
 - [Perform Impact Build for topic branches](#perform-impact-build-for-topic-branches)
 - [Perform Impact Build by providing baseline reference for the analysis of changed files](#perform-impact-build-by-providing-baseline-reference-for-the-analysis-of-changed-files)
+- [Perform Merge Build](some link)
```
drbruce-git, line 132: Need to finish the link

drbruce-git (commented):
Looks fine to me with just a reminder of a missing link in the documentation.  I am interested in he opinion of @jbyibm

BUILD.md
dennis-behm: I have updated the TOC

build.groovy
```diff
@@ -427,7 +429,11 @@ def createBuildList() {
 	// check if full build
 	if (props.fullBuild) {
 		println "** --fullBuild option selected. $action all programs for application ${props.application}"
-		buildSet = buildUtils.createFullBuildList()
+		if (repositoryClient) {
+			buildSet = buildUtils.createFullBuildList() }
+		else {
+			println "*! Full build requires a repository client connection to a DBB web application"
+		}
```
jbyibm, line 436: You are changing the current behavior here. Any reason to force providing a repository connection in case of a full build?
I very often perform a full build without repository connection just to test that my whole application can compile.

jbyibm (commented):
Can you review my comment about full build.

build.groovy
dennis-behm: Interesting use case, I have not thought about this case so far. Always interesting to see the different usages.

Doesn't your use cause an issue when zAppBuild would like to update the collection? See line 517 in build.groovy

jbyibm: Perhaps this is behavior that @drbruce-git has intentionally implemented in its design.

jbyibm (approved):
Look good

build.groovy
drbruce-git: I had always assumed that full build required a repository client so that it can store collections and build results for subsequent incremental builds.

Issue #112, opened 2021-05-25
Easytrieve language script


dennis-behm:
The DBB release 1.1.0 and the recent PTF of the DBB toolkit, allows the DBB scanner to understand Easytrieve source code and its include structure.

Make a sample easytrieve language script available as part of the zAppBuild infrastructure.

PR #154, opened 2021-11-24
Sample Easytrieve language script

dennis-behm:
The DBB scanners are able to scan Easytrieve source code to capture its' dependencies to the Easytrieve macros. (starting DBB 1.1.0 PTF2)

This sample was developed based on this sample JCL provided by the IBM Migration Utility to compile and link an Easytrieve source code:

https://www.ibm.com/docs/en/mufz/5.1?topic=jcl-jcmucl2jtwo-step-translate-link-without-proc

With little modification, it should also be able to process Easytrieve scripts.

See also #112


samples/application-conf/application.properties
```diff
@@ -125,5 +125,13 @@ testcasePgmRule =   {"category": "PROGRAMDEPENDENCY", \
               ] \
             }
 
+# Rule to locate Easytrieve mactos
```
BeckerFrank, line 128: typo. should be macros instead of mactos

dennis-behm: Thanks @BeckerFrank I have migrated my fork to support the new way of the copySourceFiles and have fixed this typo as well.

M-DLB (approved):
Thank you @dennis-behm 
Great addition to zAppBuild!

Issue #155, opened 2021-11-24
Deploy asset that don't require any transformation


marvin42-fr:
Hello,

Is it possible to deploy asset that do not require a transformation ?
For exemple we need to deploy "JCL" included in the project in a targeted HLQ.
We want to include thoses JCL in the shiplist to use with Urban Codedeploy like LOAD or DBRM.

Regards,
Eric Bartolone.

dennis-behm:
Hi Eric, can you please browse the PR, which I linked to this request? I struggle to find a good name for the new language script ... If you have a better idea than `nonBuildable`, please let me know.

marvin42-fr:
Hi Dennis,
Thanks i browse this new language script that seems to be what i'm searching for.
Does a name like 'statics' like is it for static ressources on web application?

Regards,
Eric Bartolone.

dennis-behm:
Hi Eric, I have merged the functionality via #156 to `develop` and plan to create a new release next week. Please let us know all your feedback.

marvin42-fr:
Hi Dennis, 
It's working exactly as expected for the files i need to ass to the package.
We also need to do others personilazed task on thoses files before adding them to the package (like replacing value, or create an englobing JCL) so this will help me implement it.

Thanks a lot.
Regards,
Eric Bartolone.

dennis-behm:
Hi Eric, I am closing the Issue, while it got released to zAppBuild via #160

PR #156, opened 2021-11-24
language script to copy files to dataset and report in build report

dennis-behm:
This PR relates to the reported issue #155 

There are repository files, which don't need to be built, but should be transported through the pipeline. Sample files are JCL, XML, SQL, Shared Copybooks or any type of scripts or configuration file managed in the application repository.

This contribution introduces a new language script, which 
* copies repository files to a target dataset, which is defined in a dataset mapping
* reports the files as a build output, according to the mapped deployType

It leverages the PropertyMapping API of the DBB, which was not used before in zAppBuild. Based on a PropertyMapping, it allows the application team to map a file to the build property of the target dataset. In the below sample, which is documented in `sample/application-con/file.properties` all files with the file extension `jcl` are mapped to the build property `transfer_jclPDS`, which is defined in build-conf/Transfer.properties

```
# PropertyMapping to map files using the nonBuildable language script to different target datasets 
#
# transfer_datasetMapping = transfer_jclPDS :: **/*.jcl
# transfer_datasetMapping = transfer_xmlPDS :: **/xml/*.*
```
Additionally, the application team defines the deployType for the file using a second build property mapping:

```
#
# file mapping for overwriting the default deployType of the nonbuildable language script 
#
# transfer_deployType = JCL :: **/*.jcl
# transfer_deployType = XML :: **/xml/*.*
```

### DIFF
--- a/languages/nonBuildable.groovy
+++ b/languages/nonBuildable.groovy
@@ -7,14 +7,16 @@ import com.ibm.dbb.build.report.*
 import groovy.transform.*
 
 /***
- * 
- * 
- * 
+ * Language script, which just copies files to the defined target dataset 
+ * and reports the file as a build output file in the build report
  */
 
 // define script properties
 @Field BuildProperties props = BuildProperties.getInstance()
 @Field def buildUtils= loadScript(new File("${props.zAppBuildDir}/utilities/BuildUtilities.groovy"))
+// Set to keep information about which datasets where already checked/created
+@Field HashSet<String> verifiedBuildDatasets = new HashSet<String>()
+
 @Field RepositoryClient repositoryClient
 
 println("** Building files mapped to ${this.class.getName()}.groovy script")
@@ -40,8 +42,10 @@ buildList.each { buildFile ->
 	if (targetDataset != null) {
 
 		// allocate target dataset
-		// TODO: Create a cache of targets which got allocated
-		buildUtils.createDatasets(targetDataset.split(), props.nonbuildable_srcOptions)
+		if (!verifiedBuildDatasets.contains(targetDataset)) { // using a cache not to allocate all defined datasets
+			verifiedBuildDatasets.add(targetDataset)
+			buildUtils.createDatasets(targetDataset.split(), props.nonbuildable_srcOptions)
+		}
 
 		// copy the file to the target dataset
 		String deployType = buildUtils.getDeployType("nonbuildable", buildFile, null)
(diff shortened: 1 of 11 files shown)


languages/nonBuildable.groovy
```diff
@@ -0,0 +1,77 @@
...
+@Field def buildUtils= loadScript(new File("${props.zAppBuildDir}/utilities/BuildUtilities.groovy"))
+// Set to keep information about which datasets where already checked/created
+@Field HashSet<String> verifiedBuildDatasets = new HashSet<String>()
+
+@Field RepositoryClient repositoryClient
+
+println("** Building files mapped to ${this.class.getName()}.groovy script")
+
+// verify required build properties
+buildUtils.assertBuildProperties(props.cobol_requiredBuildProperties)
```
jbyibm, line 25: Should be nonbuildable_requiredBuildProperties ?

languages/nonBuildable.groovy
```diff
@@ -0,0 +1,77 @@
...
+	
+	// obtain the target dataset based on the mapped dataset key
+	String targetDataset = props.getProperty(dsMapping.getValue(buildFile))
+	
+	if (targetDataset != null) {
+
+		// allocate target dataset
+		if (!verifiedBuildDatasets.contains(targetDataset)) { // using a cache not to allocate all defined datasets
+			verifiedBuildDatasets.add(targetDataset)
+			buildUtils.createDatasets(targetDataset.split(), props.nonbuildable_srcOptions)
```
jbyibm, line 47: I'm not sure to understand the split() here.

jbyibm (changes requested):

jbyibm:
@dennis-behm I forgot to add these comments in my review

1. For my `nonbuildable` is not "natural". But I have no good replacement. In zTrial I did something similar for IMS I used `extraFiles`.
2. Could be good to add a comment saying the files names must not be >= 8. This is natural for COBOL,COPY,BMS...

languages/nonBuildable.groovy
dennis-behm: Good catch. This should be fixed.

languages/nonBuildable.groovy
dennis-behm: Split is required to have string array ... String[] ... which is accepted by the `buildUtils.createDatasets` method

dennis-behm:
I have change the naming from `nonBuildable` to `Transfer`. @jbyibm Can you please have a final look?

jbyibm (approved):
Look good

PR #159, opened 2021-12-03
Enhancing Transfer.groovy

dennis-behm:
* Validate length of member name
* Fix the validation of the required build properties for Transfer.groovy. It requires to remove `transfer_datasetMapping` from required build properties - reason: This property is only defined as a file level property and cannot be validated as a generic build property. The file property is validated as part of the Transfer.groovy script.
* Catch Build Exceptions, like the truncation of a file


jbyibm (approved):
Look good.

PR #161, opened 2021-12-10
Store abbreviated git hash as SSI Info for load modules

dennis-behm:
This enhancement retrieves the current abbreviated git hash for changed and impacted files for the following build types: fullBuild, impactBuild and mergeBuild.

Several areas are updated in buildUtils and impactUtils to populate a PropertyMappings  object `githashBuildableFilesMap` to store a DBB build property map with these information.

The language scripts pass the abbreviated hash to the linker.

It also supports builds across multiple repositories:

image

### DIFF
--- a/build.groovy
+++ b/build.groovy
@@ -539,7 +539,7 @@ def finalizeBuildProcess(Map args) {
 			if (gitUtils.isGitDir(dir)) {
 				// store current hash
 				String key = "$hashPrefix${buildUtils.relativizePath(dir)}"
-				String currenthash = gitUtils.getCurrentGitHash(dir)
+				String currenthash = gitUtils.getCurrentGitHash(dir, false)
 				if (props.verbose) println "** Setting property $key : $currenthash"
 				buildResult.setProperty(key, currenthash)
 				// store gitUrl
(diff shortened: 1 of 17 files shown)


drbruce-git (commented):
So it looks like this change to add Git Hash SSI to load modules is not optional.  What if the customer is using the SSI for other purposes?  Shouldn't there be an option to turn this off or default is off and option to turn it on?

dennis-behm:
@drbruce-git I introduced a flag to turn it on and off. Can you please have another check?

drbruce-git (approved):

jbyibm:
@dennis-behm We may have an interest for PLUM here. 
Quick question:
- the --short=8 option is a truncation of the real githash ? or something that we can consider as an unique identifier?

dennis-behm:
> @dennis-behm We may have an interest for PLUM here. Quick question:
> 
> * the --short=8 option is a truncation of the real githash ? or something that we can consider as an unique identifier?

Hi @jbyibm 

This performs a `git rev-parse --short=8 HEAD` in the GitUtilities. It is a truncated part of the full git hash:

```
git rev-parse --short=8 HEAD                                   
5e3ccd2f

git rev-parse HEAD                  
5e3ccd2f43c18cd0dc3ba837121bcd0f50970275
```

Issue #144, opened 2021-10-22
Enhance PL/1 support with Debug and zUnit


flpet-ibm:
Enhance the PL/1 language script to support the --debug flag to compile for debug, and support build testcase programs for PL/1 zUnit test cases

dennis-behm:
Thanks Flemming for your contribution. This got released with 2.3.0 #176

PR #164, opened 2021-12-16
#144 pl/i support for debug and z unit testcase build

flpet-ibm:
Adding support for zUnit for PLI and adjust support compiling PLI programs for debug


dennis-behm:
Apologies for the additional work. I forgot to squash the commits with the previous pull request.

PR #168, opened 2022-01-11
Fix link to BUILD.md

dennis-behm:


dennis-behm:
@ijmitch Unfortunately I had merged your PR into `main`, PR from forks should be opened against `develop`, which we then bundle together to a release.

PR #169, opened 2022-01-26
update build.properties and datasets.properties with makeit environme…

BharathEluri:
…nt values


BharathEluri:
wrong create of pull request to main dbb zappbuild

PR #173, opened 2022-02-10
Support UTF-8 for user build dependency file

lburgess07:
According to [JSON Standards](https://datatracker.ietf.org/doc/html/rfc7159#section-8.1) JSON files should be encoded UTF-8. 
This PR includes support for the user build dependency file being uploaded, tagged, and parsed as UTF-8.
Note: Backwards compatibility is maintained. Previously the dependency file was being uploaded as text (converted to IBM-1047) and left untagged, and this will still be supported as described below. 

Changes Made:
- Refactored BuildUtilities.validateDependencyFile() to support and parse under three encoding/tagging scenarios:
   1. Encoded as UTF-8 and tagged as UTF-8. (JSON Standard)
   2. Encoded as IBM-1047 and tagged as IBM-1047.
   3. Encoded as IBM-1047 and untagged.  (Provides backwards compatibility)
 - Added BuildUtilities.retreiveHFSTag() which reads the file tag attribute and returns a text representation of the codepage represented by that tag. 
 - Added *.json to gitattributes to conform to JSON standards.

### DIFF
--- a/utilities/BuildUtilities.groovy
+++ b/utilities/BuildUtilities.groovy
@@ -665,7 +665,11 @@ def assertDbbBuildToolkitVersion(String currentVersion){
 	}
 }
 
-def retrieveHFSEncoding(File file) {
+/*
+ * Returns a string representation of a file's encoding calculated from its tag.
+ * 
+ */
+def retrieveHFSFileEncoding(File file) {
 	FileAttribute.Stat stat = FileAttribute.getStat(file.getAbsolutePath())
     FileAttribute.Tag tag = stat.getTag()
 	int i = 0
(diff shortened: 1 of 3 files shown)


utilities/BuildUtilities.groovy
```diff
@@ -648,6 +665,24 @@ def assertDbbBuildToolkitVersion(String currentVersion){
 	}
 }
 
+def retrieveHFSEncoding(File file) {
```
dennis-behm, line 668: Please add a description to the method. 

Additionally, I would prefer the method name as `retrieveHFSFileEncoding`

dennis-behm (changes requested):
@lburgess07 just a minor request to provide a more meaningful method name.

utilities/BuildUtilities.groovy
lburgess07: I have added a method description and changed the name as requested.

dennis-behm (approved):

PR #180, opened 2022-02-17
Formatted console output for impactResolver config only on verbose logging

dennis-behm:
The output of the impactResolver configuration for collections and searchPath should only be written to console log in case of the `--verbose` tracing option.


lburgess07:
Looks good to me!

PR #183, opened 2022-02-22
SSI info conditional via a build property for Assembler and BMS language scripts

### LINKED ISSUE #181: Make SSI documentation for Assembler and BMS optional
Adding the SSI info for the link step in language scripts Assembler and BMS is not conditional through a fileProperty like in the other language scripts.

Missing in Assembler.groovy and BMS.groovy

https://github.com/IBM/dbb-zappbuild/blob/74c662aee4a7b5a006a0c8cadd3287986823f4f4/languages/Assembler.groovy#L321-L325

https://github.com/IBM/dbb-zappbuild/blob/74c662aee4a7b5a006a0c8cadd3287986823f4f4/languages/BMS.groovy#L124-L128

This is how it should look like:
https://github.com/IBM/dbb-zappbuild/blob/74c662aee4a7b5a006a0c8cadd3287986823f4f4/languages/Cobol.groovy#L286-L291

dennis-behm:
This PR adds the conditional logic for storing the abbreviated hash via the Linker in the SSI Info field for the language scripts Assembler.groovy and BMS.groovy, where the conditional implementation was missed, when the feature was first contributed. This fixes #181.

### DIFF
--- a/languages/Assembler.groovy
+++ b/languages/Assembler.groovy
@@ -319,8 +319,11 @@ def createLinkEditCommand(String buildFile, LogicalFile logicalFile, String memb
 	String parameters = props.getFileProperty('assembler_linkEditParms', buildFile)
 
 	// obtain githash for buildfile
-	String ssi = buildUtils.getShortGitHash(buildFile)
-	if (ssi != null) parameters = parameters + ",SSI=$ssi"
+	String assembler_storeSSI = props.getFileProperty('assembler_storeSSI', buildFile)
+	if (assembler_storeSSI && assembler_storeSSI.toBoolean() && (props.mergeBuild || props.impactBuild)) {
+		String ssi = buildUtils.getShortGitHash(buildFile)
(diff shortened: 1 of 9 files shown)


jbyibm (commented):
@drbruce-git  @dennis-behm 
Can we clarify what we can expect with the `--fullBuild` option?

Today, with the `--fullBuild` I can rebuild my whole application. A possible Use Case can be a rebuild of the full application with the SSI option.  This option is also connected to the WebApp and the collection is updated when we give the `--url` option. So, why you exclude the `--fullBuild` for the SSI ?

dennis-behm:
@jbyibm In the initial implementation I only focussed on the build types impactBuild and mergeBuild and have excluded the fullBuild. Let's create another issue to add the SSI info also for fullBuild - should not be a big deal. But I don't would like to mix this with this PR.

jbyibm (approved):
look good

Issue #186, opened 2022-02-23
Capture SSI Info also in fullBuild scenario


dennis-behm:
The current implementation of the capturing of the abbreviated hash in the SSI info is only available on build types impactBuild and mergeBuild. 

This issue is about requesting the SSI Info information also in the fullBuild scenario.

dennis-behm:
Delivered

PR #187, opened 2022-02-23
Enable SSI Info for load modules in fullBuild scenario

dennis-behm:
This implements #186 to write the SSI Info also in fullBuild scenarios.


jbyibm (approved):
Look Good

PR #201, opened 2022-03-30
Leverage new SearchPathDependencyResolver and SearchPathImpactFinder API

dennis-behm:
Starting with DBB Version 1.1.2 a new set of APIs got introduced which simplifies the configuration of the dependency and impact analysis - namely these are the [SearchPathDependencyResolver](https://www.ibm.com/docs/api/v1/content/SS6T76_1.1.0/javadoc/com/ibm/dbb/dependency/SearchPathDependencyResolver.html) and the [SearchPathImpactFinder](https://www.ibm.com/docs/api/v1/content/SS6T76_1.1.0/javadoc/com/ibm/dbb/dependency/SearchPathImpactFinder.html). 

The new API allows to define filters on the file extension and the use of wildcards to specify the searchPaths. Additionally, it is capable to overwrite the file flags (isCICS, isSQL, isMQ, isDLI) based on the resolved dependencies.

The IBM Documentation for [DBB](https://www.ibm.com/docs/en/adfz/dbb/1.1.0?topic=scripts-how-manage-build-dependencies) has been updated as well including the schema to configure the new API. See sections 6 + 7.

While the previous APIs - the [DependencyResolver](https://www.ibm.com/docs/api/v1/content/SS6T76_1.1.0/javadoc/com/ibm/dbb/dependency/DependencyResolver.html) and the [ImpactResolver](https://www.ibm.com/docs/api/v1/content/SS6T76_1.1.0/javadoc/com/ibm/dbb/dependency/ImpactResolver.html) - are marked as deprecated, this contribution allows to smoothly transition to the new API. Backward compatbility is preserved. 

* Through a new build property in `application-conf/application.properties`- `useSearchConfiguration` the user is able to enable to the new APIs. 
* The sample resolution rules defined in JSON like format are converted for reference into the new *search path* configurations.
* Leveraging the new build property flag `resolveSubsystems`, the file flags of the logicalFile for middleware components  include the attributes of the resolved dependencies.
* Please note, that zAppBuild now requires at least DBB Toolkit version 1.1.2


samples/MortgageApplication/application-conf/README.md
```diff
@@ -51,8 +56,8 @@ Application properties used by zAppBuild/language/Cobol.groovy
 Property | Description | Overridable
 --- | --- | ---
 cobol_fileBuildRank | Default Cobol program build rank. Used to sort Cobol build file sub-list. Leave empty. | true
-cobol_resolutionRules | Cobol dependency resolution rules used to create a Cobol dependency resolver.  Format is a JSON array of resolution rule property keys.  Resolution rule properties are defined in `application-conf/application.properties`. | true
-cobol_compilerVersion | Default Cobol compiler version. | true
+cobol_resolutionRules | Cobol dependency resolution rules used to create a Cobol dependency resolver.  Format is a JSON array of resolution rule property keys.  Resolution rule properties are defined in `application-conf/application.properties`. ** deprecated ** | true
+cobol_dependencySearch | Cobol dependencySearch configuration to configure the SearchPathDependencyResolver. Format is a concatenated string of searchPath configurations. Strings representing the SearchPaths defined in `application-conf/application.properties`. | truecobol_compilerVersion | Default Cobol compiler version. | true
```
jbyibm, line 60: I think you have 2 lines in the same line here.

jbyibm (changes requested):
I have some comments. But it could be good to have one more reviewer.

languages/Assembler.groovy
```diff
@@ -28,15 +28,27 @@ List<String> sortedList = buildUtils.sortBuildList(argMap.buildList, 'assembler_
...
+			logicalFile = SearchPathDependencyResolver.getLogicalFile(buildFile,props.workspace)
+	} else { // use deprecated DependencyResolver
+		String rules = props.getFileProperty('assembler_resolutionRules', buildFile)
+		dependencyResolver = buildUtils.createDependencyResolver(buildFile, rules)
+		logicalFile = dependencyResolver.getLogicalFile()
+	}
+	
 	// copy build file and dependency files to data sets
-	String rules = props.getFileProperty('assembler_resolutionRules', buildFile)
-	String assembler_srcPDS = props.getFileProperty('assembler_srcPDS', buildFile)
```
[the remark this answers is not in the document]
dennis-behm: A good catch, thank you. I have replaced it with the property. The compileCommand also references this dataset.

build-conf/build.properties
```diff
@@ -45,7 +45,7 @@ applicationConfRootDir=
 #
 # Minimum required DBB ToolkitVersion to run this version of zAppBuild
 #  Build initialization process validates the DBB Toolkit Version in use and matches that against this setting 
-requiredDBBToolkitVersion=1.0.0
+requiredDBBToolkitVersion=1.1.2
```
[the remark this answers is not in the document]
dennis-behm: I now changed the implementation and conditionally load a new utility script `resolverUtilities.groovy`. I highly recommend users of zAppBuild to move to the new implementation.

dennis-behm:
Based on @jbyibm 's feedback, I have moved the usage of the new API into its own utility method, to allow backward compatibility even with older versions. I see this only as a temporary solution. It introduces some technical depth, which needs some refactoring in the future. Probably with the next DBB release, we should do some housekeeping and bump up to a full version in zAppBuild.

Users, which are then not on the required DBB toolkit level, can continue to use the zAppBuild release supporting their level of the toolkit.

@jbyibm can I ask you for another review, please?
@drbruce-git Can you also please have a look?
Thank you.

languages/LinkEdit.groovy
```diff
@@ -30,10 +30,17 @@ sortedList.each { buildFile ->
 	buildUtils.copySourceFiles(buildFile, props.linkedit_srcPDS, null, null, null)
 
 	// create mvs commands
-	String rules = props.getFileProperty('linkedit_resolutionRules', buildFile)
-	DependencyResolver dependencyResolver = buildUtils.createDependencyResolver(buildFile, rules)
-	LogicalFile logicalFile = dependencyResolver.getLogicalFile()
+	LogicalFile logicalFile
+	if (props.useSearchConfiguration && props.useSearchConfiguration.toBoolean()) { // use new SearchPathDependencyResolver
+		logicalFile = SearchPathDependencyResolver.getLogicalFile(buildFile,props.workspace)
```
jbyibm, line 35: Is SearchPathDependencyResolver exits in DBB < 1.1.2 ? If no you may have an exception.

jbyibm (changes requested):
I have just one comment in LinkEdit.groovy.

I did a quick try with DBB 1.1.0 and DBB 1.1.2 and I have this error
```
Current DBB Toolkit Version 1.1.0 does not meet the minimum required version 1.0.3. EXIT.
assert (it as int) >= ((requiredVersionList[i]) as int)
        |          |    |                  ||
        0          |    [1, 0, 3]          |2
                   false                   3
```
It looks like that the version check method does not work as expected. To succeed, I had to put 1.0.0 as minimal version for the toolkit.

drbruce-git (changes requested):
@dennis-behm  I notice in the searchpath properties you are adding the application to the search path:  Ex:
copybookSearch = search:${workspace}/?path=${application}/copybook/*.cpy 

Won't that limit the search to just single repository builds and skip dependencies in multi-repository applications?  Perhaps the searchpath's should use wildcards:
copybookSearch = search:${workspace}/?path=**/copybook/*.cpy 

Of course even here then the copybooks must be in the copybook folders in all repositories that are part of the application build.  A more versatile default search path would be
copybookSearch = search:${workspace}/?path=**/*.cpy

drbruce-git (approved):

languages/LinkEdit.groovy
dennis-behm: I have tested this with 1.0.9 . No Exception.

Issue #212, opened 2022-05-20
Simple way to overwrite build properties from the cmdline


dennis-behm:
In some occasions you would like to overwrite a specific build property. Today, the `--propFiles` infrastructure allows to read and pass options.

I am proposing to create a new cli option, which can basically overwrite any build property. An easy and simple way such as providing an overwrite like: 
`--propOverwrites key=value`  or `--propOverwrites key=value,key2=value2` as a comma separated list.

dennis-behm:
Delivered with [zApppBuild Release 2.5.0](https://github.com/IBM/dbb-zappbuild/releases/tag/2.5.0)

PR #213, opened 2022-05-23
New CLI option to set or overwrite build property

dennis-behm:
This contribution implements #212 to overwrite/set a build property from the cmdline.

Rather than using the `propFiles` option to dynamically set a build property for a particular pipeline type, this let's a pipeline scripts easily pass a (potential dynamic and varying) build property.

The cli option `propOverwrite` is processed after reading the build property files.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1421,15 +1421,15 @@ HTTP/1.1 200 OK
 
 Dependending on where are in your development process, you might want to dynamically overwrite a build property. One scenario is for example to set the `mainBuildBranch` option which is used also to clone collections and initialize the DBB collections. See [Perform Impact Build for topic branches](#perform-impact-build-for-topic-branches)
 
-To dynamically overwrite any build property, you can make use of the `--propOverwrite` cli argument. To overwrite a build property, pass it in as a key-value pair via the CLI. Several key-value pairs can be passed by comma-separating them. Please make sure that you don't run into any limitations with regards of the length of the overall build command. For mass overwrites, use the `--propFiles` option to pass in one or several property files.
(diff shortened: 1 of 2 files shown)


BUILD.md
```diff
@@ -1412,3 +1416,155 @@ HTTP/1.1 200 OK
 ```
 
 </details>
+
+### Dynamically Overwrite build properties
+
+Dependending on where are in your development process, you might want to dynamically overwrite a build property. One scenario is for example to set the `mainBuildBranch` option which is used also to clone collections and initialize the DBB collections. See [Perform Impact Build for topic branches](#perform-impact-build-for-topic-branches)
+
+To dynamically overwrite any build property, you can make use of the `--propOverwrite` cli argument. To overwrite a build property, pass it in as a key-value pair via the CLI. Several key-value pairs can be passed by comma-separating them :`--propOverwrite mainBuildBranch=develop,cobol_compilerVersion=V6`. Please note, that values which contain a comma (`,`) cannot be passed into the list. Please also make sure that you don't run into any limitations with regards of the length of the overall build command. For mass overwrites, use the `--propFiles` option to pass in one or several property files.
+
```
jbyibm, line 1425: We have some case of value with coma. It is not possible to allow escaping such values ?
`buildOrder=BMS.groovy\,Cobol.groovy`

jbyibm (commented):
Just a comment.

BUILD.md
dennis-behm: Hi Jean-Yves, I have performed multiple tests, but it did not work out - unfortunately ...

PR #215, opened 2022-05-23
Update impact resolution rule samples for static linkage scenarios

dennis-behm:
Updating the sample rules for the new SearchPath finders.


dennis-behm:
This PR was reviewed by NK and LL.

PR #216, opened 2022-05-24
Ability to manage member level build properties in specific prop files

dennis-behm:
As of today, zAppBuild manages  a hierarchy of 2 for build properties. 

Either the properties are managed globally as the defaults or are overwritten by a file property. The file property filter can be used for groups as well as individual programs.

During migration to DBB, it could be useful to manage an individual property file for each element rather than using a long string for a file property overwrite.

This PR introduces an alternate way to define a build property through a particular property member.


lauren-li (commented):
@dennis-behm I have made PR https://github.com/dennis-behm/dbb-zappbuild/pull/3 into this PR branch (`dennis-behm:propertyoverwrite`), as I wanted to make some edits to the new Build Property Management section that were too large for me to be comfortable committing without running by you first, but also somewhat difficult to explain one by one.

The proposed rewordings are based on my understanding of the information you shared with me yesterday, so please let me know if you find any inaccuracies. If the proposed suggestions are accepted, then we can also copy/paste the contents of my PR into `samples/MortgageApplication/application-conf/application.properties`.

Thank you for your hard work on this PR so far!

samples/application-conf/README.md
```diff
@@ -23,6 +23,11 @@ gitRepositoryURL | git repository URL of the application repository to establis
 excludeFileList | Files to exclude when scanning or running full build. | false
 skipImpactCalculationList | Files for which the impact analysis should be skipped in impact build | false
 jobCard | JOBCARD for JCL execs | false
+**Build Property management** | | 
+loadFileLevelProperties | Flag to enable the zAppBuild capability to load individual property files for a build file | true
+propertyFilePath | relative path to folder containing individual property files | true
+propertyFileExtension | file extension for individual property files | true
```
lauren-li, line 29: - How can the new Build Property management properties be overridden?

- We might also want to update the **file.properties** section below where it currently says:
  > All file properties for the entire application, including source files in distributed repositories of the application need to be contained either in this file or in other property files in the `application-conf` directory.
  
  Should it be updated to additionally mention the directory specified by the `propertyFilePath` property of `application-conf/application.properties`?

lauren-li (approved):
The latest version of this PR looks good to me! Thanks @dennis-behm for your hard work on this, and for your patience in explaining it to me and answering all my questions!

drbruce-git (approved):

PR #224, opened 2022-06-13
Correction PLI alternate library configuration

### LINKED ISSUE #221: PLI alternate library definition processing may be incorrect
At line 227 of PLI.groovy (in main branch), It appears that we're checking the value of the COBOL alternate library definitions property rather than the PLI value:

```
	if (props.cobol_dependenciesAlternativeLibraryNameMapping) {
		alternateLibraryNameAllocations = evaluate(props.pli_dependenciesAlternativeLibraryNameMapping)
```

I believe the conditional test should be against `pli_dependenciesAlternativeLibraryNameMapping`

dennis-behm:
`PLI.groovy` had some minor defects and was referring to cobol build properties. This PR corrects that. This closes #221


lauren-li (approved):
These updates look good to me. Thanks @dennis-behm for the fix!

PR #228, opened 2022-06-14
Correction when loading individual file properties 

dennis-behm:
During the release preparation, we noticed a problem with applications with an existing application-conf configuration that do not define the `loadFileLevelProperties` build property. This led to a NullPointerException.

The logic was extended to capture this scenario.

Multiple manual tests were performed to validate:
* `loadFileLevelProperties` not defined at all (the above scenario with an NPE).
* `loadFileLevelProperties` only set as a global property
* `loadFileLevelProperties` only defined as a file-level property
* `loadFileLevelProperties` set as a global property and as a file-level property

### DIFF
--- a/build.groovy
+++ b/build.groovy
@@ -554,8 +554,10 @@ def createBuildList() {
 	}
 	
 	// Loading file/member level properties from member specific properties files
-	println "** Populate file level properties from individual property files."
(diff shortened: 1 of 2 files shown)


lauren-li (approved):
This PR looks good to me. Thanks @dennis-behm for this fix, and for going through the test scenarios with me!

Issue #242, opened 2022-09-06
Support MQ file flag of logicalFile


dennis-behm:
Starting with DBB toolkit 1.1.1, DBB can automatically detect if a file uses MQ - See [New and Noteworthy 1.1.1](https://www.ibm.com/docs/en/dbb/1.1.0?topic=what-is-new-noteworthy#version-1-1-1) 

Enhance zAppBuild to leverage the file flag rather than relying on the `cobol_isMQ` build property.

PR #245, opened 2022-09-12
Leverage MQ file flag

dennis-behm:
Starting with DBB 1.1.1, the scanners are able to detect if a file uses MQ - see #242 

The PR implements leveraging the MQ file flag in `Assembler.groovy`, `PLI.groovy`, `Cobol.groovy` to:
* pass the necessary MQ libraries to the compilers
* generate the MQ language stub, if it is online (CICS,DLI) or batch.
* simplification of dealing with object decks and linkEditStreams, which are passed to the linkage editor.

Object decks are now always passed via a physical dataset allocation (~~initial implementation via temporary dataset allocation to the linkage editor~~).

Required include statements for the linkage editor, are passed in via a `instreamData` definition. Existing linkEditStream definitions are passed via the `instreamData` as well as the debug exit configuration.


languages/Assembler.groovy
```diff
@@ -340,14 +343,45 @@ def createLinkEditCommand(String buildFile, LogicalFile logicalFile, String memb
...
+		 linkedit.dd(new DDStatement().name("SYSIN").instreamData(sysin_linkEditInstream))
+	}
+
+	// Overwriting SYSLIN with SYSPIN, see
+	// dynalloc macro -- ddname list
+	// https://www.ibm.com/docs/en/zos/2.4.0?topic=facilities-invoking-binder-program-from-another-program
+	//
+	linkedit.dd(new DDStatement().name("SYSPIN").dsn("&&TEMPOBJ").options("SHR"))
+	if (sysin_linkEditInstream) linkedit.dd(new DDStatement().ddref("SYSIN"))
+	linkedit.setDdnames("SYSPIN,,SYSLMOD,SYSLIB,,SYSPRINT,,,,,,,,,,,,,")
```
dennis-behm, line 382: Feedback from reviewers
*  avoid using `DDNames`, which increases complexity and maintainability. This leads to moving away from temporary dataset allocations for the object deck.

languages/Assembler.groovy
```diff
@@ -340,14 +343,45 @@ def createLinkEditCommand(String buildFile, LogicalFile logicalFile, String memb
...
+	
 	linkedit.dd(new DDStatement().name("SYSPRINT").options(props.assembler_tempOptions))
 	linkedit.dd(new DDStatement().name("SYSUT1").options(props.assembler_tempOptions))
 
+	// Create linkEditInstream
+	String sysin_linkEditInstream = ''
+	// linkEdit stream specified
+	if (linkEditStream) {
+		sysin_linkEditInstream += "  " + linkEditStream.replace("\\n","\n").replace('@{member}',member)
+	} else { // dynamically add any required statements via SYSIN
```
dennis-behm, line 360: Feedback from reviewers
* This should not be conditional.

M-DLB:
Hi @dennis-behm - I reviewed the changes, and it's consistent with what we discussed on Thursday, so all good to me!

languages/Assembler.groovy
```diff
@@ -262,25 +262,20 @@ def createAssemblerCommand(String buildFile, LogicalFile logicalFile, String mem
...
-	// Pass different allocations
-	// Case: BATCH - allocation SYSIN
-	if (!buildUtils.isCICS(logicalFile) && !buildUtils.isSQL(logicalFile)) assembler.dd(new DDStatement().name("SYSIN").dsn("${assembler_srcPDS}($member)").options('shr'))
-	//	else assembler.dd(new DDStatement().name("SYSCIN").ddref("SYSIN"))
-
-	// Case: CICS - translator overwrite Ddnames
-	if (buildUtils.isCICS(logicalFile)) assembler.setDdnames("SYSLIN,,,SYSLIB,SYSPUNCH,,,,,,,,,,,,,,")
-	else if (buildUtils.isSQL(logicalFile)) assembler.setDdnames("SYSLIN,,,SYSLIB,SYSCIN,,,,,,,,,,,,,,")
+	// Pass different input allocations
+	if (buildUtils.isCICS(logicalFile)) {  	// Case: CICS - translator overwrite Ddnames
```
drbruce-git, line 266: @dennis-behm  looks like Assembler.groovy is still using DD names as part of its solution

drbruce-git (changes requested):

languages/Assembler.groovy
dennis-behm: Hi @drbruce-git Yes, this is correct. However, this is not part of the **linkEdit step** involving the MQ implementation. 

Because, the Assembler compiler does not have a  CICS or Db2 co-processor, it requires to precompile those files.
Rather than creating a new set of datasets to pass the temporary file, I would prefer to stay with the current implementation for the moment. But certainly we can address it via a different PR later on.

drbruce-git (approved):

Issue #262, opened 2022-10-21
Cleanup of application-conf sample configurations for searchPath configurations


dennis-behm:
With the builds script implementation on the branches `develop` and `main` require Dependency Based Build Version 2.0, a clean up of the sample dependency and impact rule configurations is required. 

The current version is still carrying sample definitions for both versions of dependency APIs.

Background: The `DependencyResolver` and `ImpactResolver` APIs got removed and are no longer available - see [N&N in the DBB docs](https://www.ibm.com/docs/en/dbb/2.0.0?topic=overview-what-is-new-noteworthy). Only the new searchPath configurations are valid.

Please also see the PR #201 , when the new rules got introduced in previous releases of zAppBuild. Add the new configurations to the _requiredBuildProperty_ lists.

https://github.com/IBM/dbb-zappbuild/blob/99624f601019ca6483023eb00c9d420535ced6d4/samples/application-conf/application.properties#L126-L147

dennis-behm:
Delivered via release 3.0.1

PR #267, opened 2022-10-24
Search path configuration updates

dennis-behm:
See issue #262. 

This PR is cleaning up the dependency and impact search configurations for the removed DependencyResolver and ImpactResolver APIs - see [N&N in the DBB docs](https://www.ibm.com/docs/en/dbb/2.0.0?topic=overview-what-is-new-noteworthy). 

This PR includes:
* Removal of old configurations
* Removal of the resolverUtils. Methods are now included in impactUtilities and buildUtilities
* SearchPath configurations are added to the list of required build properties
* Updated configurations for SearchPath configs for ZunitConfig.groovy
* Updated Readme


lburgess07 (commented):
I see that you are removing the `samples/MortgageApplication/application-conf/Assembler.properties` file. I imagine you are doing this, and it is safe to do so, because there aren't any ASM source files in MortgageApplication that we are building. So I'm not sure why it was in there in the first place. Just wanted to confirm this is the reason for removal?

lburgess07:
Also, since we are cleaning up old APIs, can we completely remove the BuildUtilities methods that  are commented out:
https://github.com/IBM/dbb-zappbuild/blob/develop/utilities/BuildUtilities.groovy#L308-L353
https://github.com/IBM/dbb-zappbuild/blob/develop/utilities/BuildUtilities.groovy#L689-L715
These are both methods supporting the APIs that have been removed in 2.0, which is why I commented them out, but I think we should remove the code entirely.

dennis-behm:
> I see that you are removing the `samples/MortgageApplication/application-conf/Assembler.properties` file. I imagine you are doing this, and it is safe to do so, because there aren't any ASM source files in MortgageApplication that we are building. So I'm not sure why it was in there in the first place. Just wanted to confirm this is the reason for removal?

Yes, I wanted to clean up the Mortgageapplicatio sample. While there is no ASM source, I think we can get rid of it.

lburgess07 (approved):

PR #286, opened 2022-11-28
added property group search for overrides

rsjrny:
This modification allows for ext.properties to exist along with file.ext.properties. The file.ext.properties are the highest priority, then ext.properties, then default.

I use these group properties when executing cobol.groovy and assembler.groovy to supply different parms or syslibs to groups of files.

An example I will use pertains to my assembler programs. My application consists of 400 assembler programs
I want all 400 to use Assembler.groovy but do not want to supply file.asm.properties for each individual file. to get around this I am using file extensions like
file1.asm01
file2.asm01
file3.asm02
file4.asm02
file5.asm03
file6.asm03
file7.asm04
file8.asm04

I have defined in my properties folder (a simple example)
```
asm01.properties parm=AC=1,RENT, syslib=CST550
asm02.properties parm=AC=0, syslib=CTS550
asm03.properties parm=AC=0, syslib=CTS610
asm04.properties parm= AC=1,syslib=CTS610
```

My scriptMapping has 
`dbb.scriptMapping = Assembler.groovy :: **/*.asm* `


dennis-behm:
Hi @rsjrny,

many thanks for the proposed changes to allow the grouping of build files to a set of properties. 

This is a functionality, which we had heard a few times now. At the moment zAppBuild allows global default properties and two ways to provide file level property overwrites (through DBB build properties or through the individual property files).

There are indeed several ways to accomplish this.

Rather than encoding the mapping into the directory structure or into the file extension, we are considering a mapping to a `languageDefinition` and add an additional mapping file.

@Anuprakash-Moothedath has prototyped it on his fork - I am sharing the link to the compare. 
https://github.com/IBM/dbb-zappbuild/compare/main...Anuprakash-Moothedath:dbb-zappbuild:main

It sounds like we should make that part of zAppBuild - with a proper documentation page about it.

dennis-behm:
@rsjrny As initiated some weeks ago, we were working on a similar functionality in parallel to your PR.

We made an attempt in #295 to generalise the idea you proposed - like providing a better grouping mechanism. The implementation was an easy thing, it is the about the documentation, which takes more time than expected. Would be interested in your view on it. When you have a moment, please feel free to share your feedback in OR #295 

Unfortunately, I believe, that this will overrule your PR (which has a similar idea). The community is always looking for new ideas and contributions. Please don't hesitate for a new functionality or update of an existing one.

rsjrny:
Thanks, 

I will review this option. A quick look at it and it looks like it will take care of my issue.

rsjrny:
@dennis-behm  I am happy with the #295 implementation. It will work just fine.

thanks

PR #295, opened 2022-12-13
Implement language configuration strategy to provide alternative approach to define overrides

Anuprakash-Moothedath:


docs/FilePropertyManagement.md
```diff
@@ -0,0 +1,150 @@
...
+cobol_compileParms=LIB,SOURCE :: **/cobol/epsnbrvl.cbl
+```
+
+For merging, rather than overriding, the property value of this file-level override with the default setting, you can specify the following syntax:
+
+```properties
+cobol_compileParms=${cobol_compileParms},SOURCE :: **/cobol/epsnbrvl.cbl
+```
+
+The file property path syntax also allows you to override the a build parameter for a set of files using wildcards. For example, let's assume that you are storing all CICS modules in `cobol_cics` subfolder. Using the following sample will ensure that the file flag `isCICS` is set to `true` for all COBOL files in this subfolder with the file extension `*.cbl`. However, it is recommended not to store information about the build configuration within the layout of folders, because an update would require to move files into different directories.
```
lauren-li, line 74: > because an update would require to move files into different directories.

Just to clarify, are we trying to say that we don't recommend storing build config information within subfolders because an update might require the files to be moved into different directories?

docs/FilePropertyManagement.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+
+## Default properties
+
+Default properties can be set in the corresponding language properties file.  For example, the COBOL file properties can be set in [application-conf/Cobol.properties](../samples/application-conf/Cobol.properties), while the Assembler file properties can be set in [application-conf/Assembler.properties](../samples/application-conf/Assembler.properties), and so on.
+
+zAppBuild is currently proposing to store these properties within the application repository and let the application team have control over these files. If you are looking for a more centralized way to manage the default options for all applications, you can move these definitions into the zAppBuild build framework itself by either merging them into the appropriate language property file under [build-conf](../build-conf/) or store them in a separate directory within zAppBuild itself and leverage the `applicationConfRootDir` property in [build-conf/build.propertiees](../build-conf/build.properties).
```
lauren-li, line 52: Should we say something about the order of precedence between `application-conf` vs `build-conf` here?

lauren-li (commented):
Thanks @Anuprakash-Moothedath and @dennis-behm for your hard work on this so far! I have provided copyediting suggestions at https://github.com/Anuprakash-Moothedath/dbb-zappbuild/pull/6, which include replacing the long Build Property Management comment in `application.properties` with a link to the new FilePropertyManagement documentation.

The PR https://github.com/Anuprakash-Moothedath/dbb-zappbuild/pull/6 should be checked first, but I also commented with some questions to try and clarify a couple specific sentences.

docs/FilePropertyManagement.md
dennis-behm: Hi Lauren, yes. This is correct. I rather prefer to use the folder layout to represent the functional context of the files.

Also, when moving a file to a different subfolder, dbb-zappbuild will recognise it as a changed file and process it. (which is ok).

Maybe we should reword it to:

```
However, it is not recommended to store information about the build configuration within the layout of folders. In that case, updating the build information of a file requires to move it into different directories. We rather prefer to use the folder layout to represent the functional context of the files.
```

docs/FilePropertyManagement.md
dennis-behm: Yes, we can say, that the default order of loading the properties files is `build-conf` first and `application-conf` second.

lauren-li: Just to confirm: This means that, by default, properties defined in `application-conf` would override those defined in `build-conf` - is that correct?

docs/FilePropertyManagement.md
lauren-li: Would wording it in the following way still be accurate?:
> **Note:** We do not recommend organizing the layout of repository folders/files based on build property management, because future updates to the build information of a file could require it to be moved into different folders. Instead, we recommend that the repository's folder layout represent the functional context of the files.

lauren-li: Confirmed with @dennis-behm

docs/FilePropertyManagement.md
lauren-li: Confirmed with @dennis-behm

nlopez1-ibm:
As an aid to companies who many have many LangDef Prop files, I recommend support for adding them under a subfolder of build-conf like 
   dbb-zappbuild/build-conf/langDefs/langDef01.properties 
   .                                                        /langDef02.properties  
                                                          ...

Anuprakash-Moothedath:
> 



> 

Thank you @nlopez1-ibm . We have added the subfolder langDefs under build-conf for maintaining language definition files.

drbruce-git (approved):

PR #303, opened 2022-12-19
Externalize the Timestamp Format for Build Folder and Build label

donnellt:
DCO 1.1 signed-off-by: Timothy L. Donnelly <donnellt@us.ibm.com>

Currently in zAppBuild the timestamp used to construct the build folder is in 12 hour format. Clients have complained that this 12 hour format makes it difficult to sort the build output folders in chronological order.  They would like the timestamp to be in 24 hour clock.

This update will create a new property in the "build.properties" file called `buildOutputTSformat` that will allow the client set the Date format method pattern string to what ever format they desire. The default value for this new property will be set to `yyyyMMdd.HHhmmss.mmm` to represent 24 hour format.  This only applies to the build folder timestamp.  

Within `build.groovy` a "** Build start at [timestamp]" message is emitted with a timestamp format in 12 hour format.  This will not change so as not to impact any client-side automation.

If the new `buildOutputTSformat` property is not set (or blank), then the timestamp of the build folder will be set to the hardcoded format currently found in `build.groovy`.

### DIFF
--- a/build-conf/build.properties
+++ b/build-conf/build.properties
@@ -10,7 +10,7 @@ buildPropFiles=datasets.properties,dependencyReport.properties,Assembler.propert
 buildListFileExt=txt
 
 #
-# Service URL for the Git provider to have a visual comparision of two hashes  
+# Service URL for the Git provider to have a visual comparision of two hashes
 # Leveraged as a build result property <props.gitRepositoryURL>/compare/
 # samples: GitHub : /compare/ ; GitLab :  /-/compare/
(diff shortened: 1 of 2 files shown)


donnellt:
My apologies for not assigning this pull request to myself.  I tried to update it after it was submitted, but GITHUB will not allow me.

dennis-behm:
@donnellt Thank you for this contribution.

This timestamp format does not only apply to the build output folder, but it is also setting the label within the build result - which I believe is a good thing.

Thank you for the contribution!

PR #312, opened 2023-01-12
Externalize the Build Folder Timestamp Format

donnellt:
Signed-off-by: Timothy L. Donnelly <donnellt@us.ibm.com>

Currently in zAppBuild the timestamp used to construct the build folder is in 12 hour format. Clients have complained that this 12 hour format makes it difficult to sort the build output folders in chronological order. They would like the timestamp to be in 24 hour clock.

This update will create a new property in the "build.properties" file called buildOutputTSformat that will allow the client set the Date format method pattern string to what ever format they desire. The default value for this new property will be set to yyyyMMdd.HHhmmss.mmm to represent 24 hour format. This only applies to the build folder timestamp and build results entry name.

Within build.groovy a "** Build start at [timestamp]" message is emitted with a timestamp format in 12 hour format. This will not change so as not to impact any client-side automation.

If the new buildOutputTSformat property is not set (or blank), then the timestamp of the build folder will be set to the hardcoded format currently found in build.groovy.

### DIFF
--- a/build-conf/build.properties
+++ b/build-conf/build.properties
@@ -10,7 +10,7 @@ buildPropFiles=datasets.properties,dependencyReport.properties,Assembler.propert
 buildListFileExt=txt
 
 #
-# Service URL for the Git provider to have a visual comparision of two hashes  
+# Service URL for the Git provider to have a visual comparision of two hashes
 # Leveraged as a build result property <props.gitRepositoryURL>/compare/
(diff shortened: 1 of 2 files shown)


dennis-behm (approved):
Thank you for the contribution!

dennis-behm:
Apologies for the inconvenience for additional work with the sign-off of the commits

PR #320, opened 2023-02-01
Fixed lib imports and deployType in REXX.groovy

M-DLB:
Signed-off-by: M-DLB <mathieu.dalbin@fr.ibm.com>


dennis-behm (approved):
Thanks for fixing the calculation of the deployType

Issue #332, opened 2023-03-16
Cobol.groovy is using the wrong output dataset definition for scanning the load module


dennis-behm:
When scanning the output module - instead of `props.linkedit_loadPDS` this needs to be `props.linkedit_cobolPDS`.

So far this was no problem, because both dataset definitions were using the same last level qualifier.

https://github.com/IBM/dbb-zappbuild/blob/5a649f0a591625be0bfe39dd880ef1d43ba8272f/languages/Cobol.groovy#L106

dennis-behm:
Delivered via zAppBuild Release 3.2.0 (#354)

PR #334, opened 2023-03-16
Fix load libraries for scanning for link dependencies

dennis-behm:
This is resolving #332


M-DLB (approved):
All looks good to me.
Thanks @dennis-behm

Issue #340, opened 2023-03-29
Defect with reporting of external impacts 


dennis-behm:
DBB 2.0 has changed the APIs to get all impacted files. The functionality is currently broken and needs to be refactored.

dennis-behm:
Delivered in [zAppBuild Release 3.2.0](https://github.com/IBM/dbb-zappbuild/releases/tag/3.2.0)

PR #341, opened 2023-03-29
Refactor zAppBuild reporting capabilities

dennis-behm:
Reporting capabilities in zAppBuild were previously managed in the ImpactUtilities script. While this is a feature of zAppBuild, it does not need to be part of the ImpactUtils script.

This PR is
* moving the the methods of the reporting capabilities to its dedicated utility script and addresses #340 .
* some helper methods got moved to the build utilities rather than having them in the impact utilities.

### DIFF
--- a/utilities/ReportingUtilities.groovy
+++ b/utilities/ReportingUtilities.groovy
@@ -150,17 +150,23 @@ def calculateLogicalImpactedFiles(List<String> fileList, Set<String> changedFile
 			if(matchesPattern(cName,collectionMatcherPatterns)){ // find matching collection names
 
 				def Set<String> externalImpactList = collectionImpactsSetMap.get(cName) ?: new HashSet<String>()
-				// query dbb web app for files with all logicalDependencies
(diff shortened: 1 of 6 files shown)


utilities/ReportingUtilities.groovy
```diff
@@ -0,0 +1,356 @@
...
+	if(logicalDependencies.size != 0) {
+
+		// iterate over collections
+		metadataStore.getCollections().each{ collection ->
+			String cName = collection.getName()
+			if(matchesPattern(cName,collectionMatcherPatterns)){ // find matching collection names
+
+				def Set<String> externalImpactList = collectionImpactsSetMap.get(cName) ?: new HashSet<String>()
+				// query dbb metadatastore for files with all logicalDependencies
+				def logicalImpactedFilesCollections = metadataStore.getImpactedFiles([cName], logicalDependencies);
```
dennis-behm, line 158: Rather than querying for each individual collection name, pass the list of collections for which the filter applies.

M-DLB (approved):
That looks good to me!
Thanks @dennis-behm

PR #348, opened 2023-04-27
Adds CICS Resource builder build capability to zAppbuild

odttlnt:
This request extends the functionality of the build by allowing users to build CICS resource definitions in yaml format. The changes will allow the CICS yaml formatted file to be build by the CICS Resource builder tool. It will generate an artefact which will be in the legacy CSD format. This artefact can then be used in a DFHCSDUP job to apply the changes to the CICS region. The PR is a part of the Infrastructure as code effort. 
The PR consists of CRB. groovy file for actual build and its corresponding properties files CRB.properties


odttlnt:
Somethings not right

Issue #367, opened 2023-06-14
Enhancement in zUnit processing


dennis-behm:
Customer feedback/scenario:
* Would like to run the zUnit test cases on feature branches and have a dedicated quality pipeline (for run multiple quality checks)
* Don't would like to build or process the zUnit test cases on the releases branches.

Problem statement: When creating the feature branches from the release branch, which will clone the dependency information (collections), the dependency information is incomplete and misses the test case dependency, which is inserted via the `createTestcaseDependency' option.

The desired state would be for the release builds:
 
1. Scan generated test case (aka record a logical file), but don’t build it, because it is not necessary.
1. Scan the bzucfg file (aka record a logical file), but don’t process it (because it is part of the quality control build and not the release build).

The above bullets need to be met to allow the createTestCaseDependency option to insert the custom dependency.

To implement 1) I plan to skip building the generated test case (in Cobol.groovy) based on the cli option `runzTests` 
2) is already given with the latest update for #308 .

PR #368, opened 2023-06-16
Ability to maintain consistent metadata store for dependencies on generated zunit test case

dennis-behm:
This is implementing the scenario outlined in #367 to allow users of zAppBuild to

* scan the generated zunit test case configuration,
* add the custom dependency via `createTestCaseDependency` configuration,
* but skip building the generated test case

to maintain consistent dependency information for zunit test setups for any feature/topic branch builds, which clone the dependency information to allow to automatically identify the generated test case via impact analysis, when the application program is changed.


M-DLB (approved):
Approved - Thank you @dennis-behm

PR #369, opened 2023-06-19
Use IDENTIFY to sign load modules

dennis-behm:
This is enhancing the traceability capabilities between source code and the produced executables.

It leverages the [IDENTIFY statement](https://www.ibm.com/docs/en/zos/2.5.0?topic=reference-identify-statement) of the Binder/linkage-editor infrastructure to generate a string identifying the **owning application** and the **state of the git repository**, when the output was produced.

The binder is adding these information into the user data field of the load module, which can be retrieved with amblist. 

```
          DATE         USER DATA
CSECT:    EPSCMORT
          06/19/2023   MortgageApplication/a21b6ab0
```

The feature is available for build files mapped to mapped to `Assembler.groovy`, `Cobol.groovy` and `PLI.groovy`. Because the signing it based on the git hash, it is only available for pipeline builds.


utilities/BuildUtilities.groovy
```diff
@@ -822,6 +822,46 @@ def matches(String file, List<PathMatcher> pathMatchers) {
...
+def generateIdentifyStatement(String buildFile) {
+
+	def String identifyStmt
+
+	if((props.mergeBuild || props.impactBuild || props.fullBuild) && MetadataStoreFactory.getMetadataStore() != null) {
+		
+		String member = CopyToPDS.createMemberName(buildFile)
+		String identifyString = props.application + "/" + getShortGitHash(buildFile)
+		//   IDENTIFY EPSCSMRT('MortgageApplication/abcabcabc')
+		identifyStmt = "  " + "IDENTIFY ${member}(\'$identifyString\')"
```
Crosswind, line 850: It is generally a good idea to also include some *eyecatcher* text that makes this information easier to find when browsing the load module (in contrast to using `amblist`):
`identifyStmt = "  " + "IDENTIFY ${member}(git-hash: \'$identifyString\')"`

That way you can search for git-hash and will always have the information visible even when browsing the load module.

Crosswind (changes requested):
I would recommend a small enhancement that will make it easier to find this information even without using `amblist`

docs/HOWTOS.md
```diff
@@ -0,0 +1,84 @@
...
+
+It leverages the [IDENTIFY statement](https://www.ibm.com/docs/en/zos/2.5.0?topic=reference-identify-statement) of the linkage editor.
+
+**How to enable the signing**
+
+The feature is enabled by default and is controlled via the `assembler_identifyLoad`, `cobol_identifyLoad` and the `cobol_identifyLoad` properties managed the corresponding property files of the language scripts.
+
+**What to expect**
+
+The feature, will generate the linker statement `IDENTIFY` for the build file. It follows the structure:
```
M-DLB, line 21: Can you remove the extra comma after feature?

docs/HOWTOS.md
```diff
@@ -0,0 +1,84 @@
...
+
+This page collects information to configure and use specific features of zAppBuild. Available How-tos:
+
+- [How-to and Frequently Asked Questions](#how-to-and-frequently-asked-questions)
+  - [Signing load modules and program objects](#signing-load-modules-and-program-objects)
+
+
+## Signing load modules and program objects
+
+zAppBuild is configured to automatically insert a reference of the version of the source code (the githash), which was used to build the executables. At the moment, the feature is only available for build files which are mapped to `Assembler.groovy`, `Cobol.groovy` and `PLI.groovy`. It is available for pipeline builds.
```
M-DLB, line 11: I would rephrase to: "zAppBuild can be configured....."

docs/HOWTOS.md
```diff
@@ -0,0 +1,84 @@
...
+
+It leverages the [IDENTIFY statement](https://www.ibm.com/docs/en/zos/2.5.0?topic=reference-identify-statement) of the linkage editor.
+
+**How to enable the signing**
+
+The feature is enabled by default and is controlled via the `assembler_identifyLoad`, `cobol_identifyLoad` and the `cobol_identifyLoad` properties managed the corresponding property files of the language scripts.
```
M-DLB, line 17: I believe a word is missing at the end of the sentence: "managed by" or "managed in" ?

docs/HOWTOS.md
```diff
@@ -0,0 +1,84 @@
...
+1                         *** M O D U L E  M A P ***
+...
+```
+
+**Known limitations**
+* It does not generate the information for link cards, that are managed in the git repository
+* It does not manage scenarios where multiple object decks are assembled into the load module
+
+
+**When is this useful**
```
M-DLB, line 56: Maybe add a question mark '?' at the end of the sub-title?

docs/HOWTOS.md
```diff
@@ -0,0 +1,84 @@
...
+
+If you want to quickly understand the version of the source code which is running in a runtime, you can use the [amblist service aid](https://www.ibm.com/docs/en/zos/2.5.0?topic=sets-amblist-service-aid) to retrieve the information for a load module to answer which application system the modules belongs to and which version of the code is currently executed.
```
M-DLB, line 58: I would split in two sentences: ".... to retrieve the information for a load module or a program object. The IDENTIFY information can help you to understand to which application the module belongs and which version of the source code was used to build the executed module".

docs/HOWTOS.md
```diff
@@ -0,0 +1,84 @@
...
+
+Using the sample JCL, helps to obtain the information
```
M-DLB, line 60: "The following sample JCL showcases the use of AMBLIST to display load modules' or program objects' information:"

docs/README.md
```diff
@@ -6,5 +6,6 @@ This folder contains supplemental documentation to help explain the implementati
 |-|-|
 |[Build Management](BUILD.md)|Documentation of the main and start build script `build.groovy`, reference of the command line options of the build script and documentation of the build scenarios supported by zAppBuild |
 |[Build Property Management](FilePropertyManagement.md)|Documentation on managing default and overriding build properties for specific application files|
+|[Frequently Asked Questions](HOWTOS.md)|Documentation on recipes how to use individual functions of zAppBuild|
```
M-DLB, line 9: "Recipes about the use of specific zAppBuild's features"

samples/application-conf/Assembler.properties
```diff
@@ -56,6 +56,16 @@ assembler_dependencySearch=${asmMacroSearch}${asmCopySearch}
...
 
+#
+# flag to generate IDENTIFY statement during link edit phase
+# to create an user data record (IDRU) to "sign" the load module with 
+# an identify String: <application>/<abbreviatedGitHash>
+# to increase traceability  
+#
+# can be overridden by file properties 
+# default: true
+cobol_identifyLoad=true
```
M-DLB, line 67: Shouldn't it be ```assembler_identifyLoad``` ?

utilities/BuildUtilities.groovy
M-DLB: Good point, I concur to @Crosswind 's point :)

M-DLB (commented):
@dennis-behm - Reviewed it and overall looks pretty good! Thank you

dennis-behm:
@M-DLB, thanks for the review. I have incorporated your feedback.

@Crosswind, @M-DLB,
Still one remaining item - the eye catcher. Instead of `git-hash` as the eye catcher - I would propose to use `scmRef`, while the git hash is just one part of the information. Thoughts?

M-DLB:
> @Crosswind, @M-DLB, Still one remaining item - the eye catcher. Instead of `git-hash` as the eye catcher - I would propose to use `scmRef`, while the git hash is just one part of the information. Thoughts?

Why not using comma-separated key-value pairs, like ```SCMRef:yyyyy,GitHash:xxxxx```, where yyyyy is the Application name (or the owning Git repository) and xxxxx the Git hash?

M-DLB:
Also, how do you deal with the 40-character limitation? (I believe you told me 40 was the limit)
Would the record be truncated by the binder?

Crosswind:
> Also, how do you deal with the 40-character limitation? (I believe you told me 40 was the limit)
> Would the record be truncated by the binder?

This is handled in the code @dennis-behm wrote. It won’t return a value if it exceeds 40 characters.

dennis-behm:
The implementation today provides `<application>/<abbrev.GitHash>` as the identifier without an eye catcher.

Instead of the proposed `scmRef`, we could just `SCM:` as the prefix:
```
SCM:MortgageApplication/abcabcabc
```

@M-DLB, there is a [warning message being issued](https://github.com/IBM/dbb-zappbuild/pull/369/files#diff-f928750f9bf48070fb80da9a0275759f38ecc1331b035f33b307d721e6b2dd96R853) into the log, and I assume that the LinkageEditor would write a warning as well.

Btw. the [IDENTIFY option](https://www.ibm.com/docs/en/zos/2.5.0?topic=reference-identify-statement) allows 80 chars for program objects! So we could debate if the warning should occur on 40 or 80 characters.

M-DLB:
About the 40/80 bytes limitation: ideally, you should check if the target dataset is a PDS or a PDS-E so you know if the executable is load module or a program object. Depending on the check result, you may allow 40 chars or 80 chars. For a simple management, you can limit it 40 chars, but are we sure that it doesn't get to the limit easily?

If the short Git hash is 7 chars (could be up to 12 according to [Git Doc](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection)), and with the 'SCM:' and '/' strings to split the information, it leaves 28 chars for the application name for load modules.

JZOS provides a way to check for the DSORG: https://www.ibm.com/docs/api/v1/content/SSYKE2_8.0.0/com.ibm.java.zsecurity.api.80.doc/com.ibm.jzos/com/ibm/jzos/ZFile.html#getDsorg--

M-DLB (approved):

Issue #370, opened 2023-06-20
Build file still scanned in impact analysis phase


dennis-behm:
With the contribution of #308, we noticed a case, that the file is still scanned as part of the impact analysis phase.

Usually, the SearchPathImpactFinder will leverage the internal cache of logical files Because the file is not know in the cache, the file will be rescanned. So, It requires to register the "unscanned" LogicalFile to the cache. 

https://github.com/IBM/dbb-zappbuild/blob/5000f02e312e99a6488a2218a84d65d05bdd8180/utilities/ImpactUtilities.groovy#L219

dennis-behm:
Delivered in [zAppBuild Release 3.3.0](https://github.com/IBM/dbb-zappbuild/releases/tag/3.3.0)

PR #371, opened 2023-06-21
Avoid additional scan of build file during impact analysis 

dennis-behm:
This is implementing #370. To avoid an additional scan of the build file during impact analysis phase, the logical file is registered in the internal cache of DBB.


M-DLB (approved):
All good - Thank you @dennis-behm

PR #375, opened 2023-07-07
Support for zCEE OpenAPI 3.0 processing

M-DLB:
The zCEE3.groovy script provides support for z/OS Connect builds using gradle.

This zCEE3.groovy is aiming to support the OpenAPI 3.0 compatible API definitions (as created through the z/OS Connect Designer). The script uses the gradle configuration available in the OpenAPI definition to create a WAR file that contains the API's deployable artifacts.

Sample properties files are provided, in the ```build-conf``` configuration folder, for enterprise-wide properties, and in ```application-conf``` folder for application-level overrides if necessary. File-level overrides can also be implemented in ```file.properties```.


samples/application-conf/README.md
```diff
@@ -267,6 +267,16 @@ zunit_CodeCoverageHost | Headless Code Coverage Collector host (if not specified
 zunit_CodeCoveragePort | Headless Code Coverage Collector port (if not specified IDz will be used for reporting) | true 
 zunit_CodeCoverageOptions | Headless Code Coverage Collector Options | true
 
+### zCEE3.Properties
```
dennis-behm, line 270: @M-DLB can you please align to the standard language prefix, which is the base name of the language script in lower case?

build-conf/README.md
```diff
@@ -292,6 +292,16 @@ zunit_reportDatasets | Comma separated list of 'report' type data sets
 zunit_reportOptions | BPXWDYN creation options for creating 'report' type data sets
 zunit_dependenciesDatasetMapping | DBB property mapping to map dependencies to different target datasets
 
+### zCEE3.Properties
```
dennis-behm, line 295: @M-DLB can you please align to the standard language prefix, which is the base name of the language script in lower case?

build-conf/zCEE3.properties
```diff
@@ -0,0 +1,25 @@
...
+# When set to false, will pick the existing WAR file if found or will build it if no WAR file is found
+zCEE_alwaysRebuildWAR=true
+
+#
+# Shell environment used to run the gradle command
+zCEE_shellEnvironment=bash
+
+#
+# Path to gradle executable
+zCEE_gradlePath=/var/gradle/gradle-7.6/bin/gradle
```
dennis-behm, line 20: Let's add this into the description of the property as a sample, but keep the definition clean.

languages/zCEE3.groovy
```diff
@@ -0,0 +1,96 @@
...
+	if (props.verbose) 
+		println("** Expected path to api.war: ${WarLocation}")
+	File WarFile = new File(WarLocation)
+	String[] command;
+	String commandString;
+	
+	if (alwaysRebuildWAR || !WarFile.exists()) {
+		println("** File api.war doesn't exist for this z/OS Connect project. Building it...")
+		String JAVA_OPTS = props.getFileProperty('zCEE_gradle_JAVA_OPTS', buildFile)
+		String gradlePath = props.getFileProperty('zCEE_gradlePath', buildFile)
```
dennis-behm, line 43: Please add a check if gradle exists at the provided path.

build-conf/zCEE3.properties
```diff
@@ -0,0 +1,25 @@
...
+
+#
+# Comma separated list of required build properties for zCEE3.groovy
+zCEE_requiredBuildProperties=zCEE_alwaysRebuildWAR, zCEE_shellEnvironment,\
+                                zCEE_gradlePath, zCEE_gradle_JAVA_OPTS
+
+
+#
+# When set to true, always rebuild the WAR file, even if one is found in the build folder of the z/OS Connect project folder.
+# When set to false, will pick the existing WAR file if found or will build it if no WAR file is found
```
dennis-behm, line 11: A WAR file should not be under version control at all. I think it is safe to assume, that we always need to produce a new *war file ZCEE3 anyway.

My suggestion would be to drop this property, and we always build when the openapi specification has changed.

languages/zCEE3.groovy
```diff
@@ -0,0 +1,96 @@
...
+		StringBuffer shellOutput = new StringBuffer()
+		StringBuffer shellError = new StringBuffer()
+	
+		ProcessBuilder cmd = new ProcessBuilder(shellEnvironment, gradlePath, gradleBuildLocation);
+		Map<String, String> env = cmd.environment();
+		env.put("JAVA_OPTS", JAVA_OPTS);
+		cmd.directory(new File(workingDir));
+		Process process = cmd.start()
+		process.consumeProcessOutput(shellOutput, shellError)
+		process.waitFor()
```
dennis-behm, line 59: Can we capture the output of the gradle build in a log file similar to the CICS Resource Builder script, and make sure we store it in the output directory?

https://github.com/IBM/dbb-zappbuild/blob/1996c0436b72787c9e9257defc91fbfc0c7e6add/languages/CRB.groovy#L75-L79

dennis-behm (changes requested):
@M-DLB , I have added a couple of comments and wished in the below review.

build-conf/README.md
M-DLB: Done

build-conf/zCEE3.properties
M-DLB: Done

languages/zCEE3.groovy
M-DLB: Done

languages/zCEE3.groovy
M-DLB: Done

build-conf/zCEE3.properties
M-DLB: I completely agree, but it's the z/OS Connect Designer that creates the WAR file, and store it inside the Git repository (and there is nothing you can do to avoid it as far as I know). 
I think we should still propose this option, but set it as optional, and set the default to true (as I did in this version).

dennis-behm: Can we make a comment about using `.gitignore` to avoid storing the WAR there?

samples/application-conf/zCEE3.properties
```diff
@@ -2,24 +2,26 @@
...
-                                zCEE_gradlePath, zCEE_gradle_JAVA_OPTS
+zcee3_requiredBuildProperties=zcee3_shellEnvironment, zcee3_gradlePath, zcee3_gradle_JAVA_OPTS
 
 
 #
 # When set to true, always rebuild the WAR file, even if one is found in the build folder of the z/OS Connect project folder.
 # When set to false, will pick the existing WAR file if found or will build it if no WAR file is found
-# zCEE_alwaysRebuildWAR=true
+# When not specified, the default value is set to true
+#zcee3_alwaysRebuildWAR=
```
dennis-behm, line 12: @M-DLB Still believe, we should not offer the property at all, because someone could just build it somewhere (locally), then inject the WAR into the repo and nobody knows what the input to this WAR was.

M-DLB: Fine, and I tend to agree! Completely removed :)

build-conf/zCEE3.properties
```diff
@@ -0,0 +1,20 @@
...
+#
+# Comma separated list of required build properties for zCEE3.groovy
+zcee3_requiredBuildProperties=zcee3_shellEnvironment, zcee3_gradlePath, zcee3_gradle_JAVA_OPTS
+
+#
+# Shell environment used to run the gradle command
+zcee3_shellEnvironment=bash
+
+#
+# Absolute path to gradle executable
```
dennis-behm, line 12: Can we add a comment, that this is gradle on z/OS UNIX System Services? The script does not support a "distributed" build.

languages/zCEE3.groovy
```diff
@@ -0,0 +1,110 @@
...
+    File logFile = new File("${props.buildOutDir}/${member}.zCEE3.log")
+    if (logFile.exists())
+        logFile.delete()
+    
+    String JAVA_OPTS = props.getFileProperty('zcee3_gradle_JAVA_OPTS', buildFile)
+    String gradlePath = props.getFileProperty('zcee3_gradlePath', buildFile)
+
+    File gradleExecutable = new File(gradlePath)
+    if (!gradleExecutable.exists()) {
+        def errorMsg = "gradle wasn't find at location $gradlePath."
```
dennis-behm, line 49: Can we prefix the msg with `*!`, please?

samples/application-conf/zCEE3.properties
```diff
@@ -0,0 +1,20 @@
+# Releng properties used by language/zCEE3.groovy
+
+#
+# Comma separated list of required build properties for zCEE3.groovy
```
dennis-behm, line 4: The required build properties should not be part of the `application-conf` folder.

What I noticed - all these settings actually are not really application specific, are they? They are a copy of the global configuration. I don't think, that we need to provide a sample in `application-conf`, and if we want then you missed this file specified on the `applicationPropFiles` list.

build-conf/zCEE3.properties
```diff
@@ -0,0 +1,20 @@
+# Releng properties used by language/zCEE3.groovy
+
+#
+# Comma separated list of required build properties for zCEE3.groovy
```
dennis-behm, line 4: You are missing to add the properties file to the `buildPropFiles` list in `build-conf/build.properties`

dennis-behm (commented):
Hi @M-DLB some minor comments that are worth to consider

build-conf/README.md
```diff
@@ -292,6 +292,15 @@ zunit_reportDatasets | Comma separated list of 'report' type data sets
...
```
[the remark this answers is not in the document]
M-DLB: Done

samples/application-conf/zCEE3.properties
M-DLB: Property files removed from the samples

build-conf/zCEE3.properties
M-DLB: Added now

dennis-behm (approved):
Thanks for the new language processing script!

PR #377, opened 2023-07-07
Added support for CICS Resource Builder

M-DLB:
This PR is to add support for the CICS Resource Builder utility, to help building CICS configuration from YAML files.
The language script is called ```CRB.groovy``` and you can have property files, at the enterprise-level in the ```build-conf``` folder, or at the application-level in ```application-conf``` subfolder of the application's workspace, with overrides in ```file.properties```.


dennis-behm (approved):
@M-DLB, thank you for the contribution to demonstrate how infrastructure as code can be included into the build framework zAppBuild.

PR #385, opened 2023-07-11
Tolerate old dbb.scannerMapping configurations

dennis-behm:
The development branch contains the enhancement #308, which implementation would stop a build, when a malformed `dbb.scannerMapping` configuration is detected due to the assertion exception.

The same property name was used previously to map to the ZUnitConfigScanner and is now extended to specify all the scanner mappings - see the defaults in build-conf/defaultzAppBuildConf.properties.

While we could meet some old application-confs with an old `dbb.scannerMapping`, this is now reported as a warning and skipped. It does not fail the build.

(redo of PR #383)

zAppBuild Testframework output:

[dennis-behm-381_tolerate_old_dbb_scriptmappings.log](https://github.com/IBM/dbb-zappbuild/files/12016533/dennis-behm-381_tolerate_old_dbb_scriptmappings.log)

### DIFF
--- a/utilities/DependencyScannerUtilities.groovy
+++ b/utilities/DependencyScannerUtilities.groovy
@@ -11,6 +11,9 @@ import groovy.json.JsonSlurper
 // define script properties
 @Field BuildProperties props = BuildProperties.getInstance()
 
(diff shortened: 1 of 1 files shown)


M-DLB (approved):
Approved - Thanks @dennis-behm

PR #408, opened 2023-08-29
Updated MortgageApp CICS resources

M-DLB:
Updates to the CICS Resources definition files to be used with CICS Resource builder, to include a missing program.


dennis-behm (approved):
Thanks @M-DLB

Issue #404, opened 2023-08-24
Time format "yyyyMMdd.HHmmss.mmm" or "yyyyMMdd.HHmmss.SSS" ?


FALLAI-Denis:
Hi,

Many zAppbuild scripts use the formatting string `yyyyMMdd.HHmmss.mmm` to format a date-time, but shouldn't the string `yyyyMMdd.HHmmss.SSS` be used instead?

`mmm` does not stand for "microsecond", unlike `SSS`, (`S` for fraction of second).

Confer [SimpleDateFormat patterns](https://groovy.apache.org/blog/groovy-dates-and-times-cheat#_simpledateformat_patterns).

M-DLB:
Yes, I think you're right. I never noticed that before...
In some examples I've seen on our environment: 
`Build start at 20230911.043529.035`
The last part `035` is actually the minutes on 3 digits, as we see the same minutes in `043529`. I've check on other builds and it's always the same... 
We'll work on a fix for the next release! 

Thank for reporting that problem @FALLAI-Denis !

PR #425, opened 2023-10-06
Fixing timestamp issue and adding decoration

M-DLB:
This PR is fixing the problem reported [here](https://github.com/IBM/dbb-zappbuild/issues/404), plus additional issues (error when displaying the list of input arguments, decoration to identify the different steps of the build process).

### DIFF
--- a/build.groovy
+++ b/build.groovy
@@ -27,7 +27,7 @@ import groovy.cli.commons.*
 @Field startTime = new Date()
 
 // start time message
-props.startTime = startTime.format("yyyyMMdd.hhmmss.mmm")
+props.startTime = startTime.format("yyyyMMdd.hhmmss.SSS")
 println("\n** Build start at $props.startTime")
 
 // initialize build
@@ -89,7 +89,7 @@ if (props.error)
 //********************************************************************
 
 def initializeBuildProcess(String[] args) {
-	if (props.verbose) println "** Initializing build process . . ."
+    println "**************** Initialization of the build process ****************"
 
 	def opts = parseArgs(args) // parse incoming options and arguments
 	populateBuildProperties(opts) // build properties initial set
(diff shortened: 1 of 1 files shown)


dennis-behm:
@M-DLB Thank you for the updates.

https://github.com/M-DLB/dbb-zappbuild/blob/Fix-Timestamp/build-conf/build.properties#L108 defines also a valid time stamp. It is a bit of a "chicken and the egg" issue here, I guess, if we use the hard-coded vs the property-defined value.

Anyway, should we move to a 24-h format generally to be more consistent? We could simplify the code and remove the property.

M-DLB:
I was not aware of this property! I think it makes sense to remove it, and only use the `props.startTime` to generate the build label. This would remove the `buildOutputTSformat` property.
What do you mean by "24-h format generally to be more consistent"?

dennis-behm:
Hi @M-DLB ,

> What do you mean by "24-h format generally to be more consistent"?

The **buildOutputTSformat** has been added to use the 24 hour format `yyyyMMdd.HHmmss.mmm` for the build output directory, that when you sort the build output directory or the build label, it provides the correct chronological order. Lower case `hh` in `yyyyMMdd.hhmmss.mmm` is the 12 hour format, leading to confusion.

I would rather go for the 24 format for all, then :-)

build-conf/README.md
```diff
@@ -41,7 +41,7 @@ applicationDefaultPropFiles | Comma separated list of default application confi
 buildListFileExt | File extension that indicates the build file is really a build list.
 applicationConfRootDir | Alternate root directory for application-conf location.  Allows for the deployment of the application-conf directories to a static location.  Defaults to ${workspace}/${application}
 createBuildOutputSubfolder | Option to create a subfolder with the build label within the build output dir (outDir). Default: true.
-buildOutputTSformat | Defines the build timestamp format for build output subfolder and build label.
+(Deprecated) buildOutputTSformat | Defines the build timestamp format for build output subfolder and build label.
```
dennis-behm, line 44: @M-DLB Let's completely drop the `buildOutputTSformat` property. It is no longer used in your implementation anyway.

M-DLB: Alright, but it should be part of the changelog that this property no longer exists.

dennis-behm (approved):
@M-DLB Thanks for the updates. This makes it more consistent. A info message needs to be added to the release notes, that the timestamp now leverages the 24-h format for labels of the build result, the build directory (if used) etc.

PR #429, opened 2023-10-16
Fix PL/I requiredBuildProperties property name

lauren-li:
Correct the property name for PL/I requiredBuildProperties so that it is applied when the [PL/I language script](https://github.com/IBM/dbb-zappbuild/blob/main/languages/PLI.groovy) runs.


dennis-behm (approved):
Thank you @lauren-li

Issue #432, opened 2023-10-17
Pipeline builds with Git detached head scenarios only returning the last segment of branch names


dennis-behm:
Computation of the git branch in detached head scenario, is returning only last segment of the reference

https://github.com/IBM/dbb-zappbuild/blob/d018b5c83cfb9d2a1f3f40c82aa8e2c8a05a9821/utilities/GitUtilities.groovy#L56-L83

If following a branch naming convention with multiple segments, like below, this can cause conflicts with the DBB collection and build groups. 

```
refs/remotes/origin/feature/setmainbuildbranch
refs/remotes/origin/hotfix/rel-1.0.0/myfix
refs/remotes/origin/implementAI/myFeatureImpl
```

dennis-behm:
Delivered via [zAppBuild Release 3.5.0](https://github.com/IBM/dbb-zappbuild/releases/tag/3.5.0)

Issue #143, opened 2021-10-08
Wrong git branch name returned in detached head situation


dennis-behm:
With some pipeline orchestrators, the checkout to build workspace, leads to a detached Head situation. 

zAppBuild can detect a `detached head` situation, and obtains the branch name through `git show`, see:

https://github.com/IBM/dbb-zappbuild/blob/8971bed1b701bb017d8b6a88bfe67ed2a70b61e8/utilities/GitUtilities.groovy#L57-L58

However, the current logic returns the _last_ reference from the command.

https://github.com/IBM/dbb-zappbuild/blob/8971bed1b701bb017d8b6a88bfe67ed2a70b61e8/utilities/GitUtilities.groovy#L70-L80

In my situation, the checked out codebase is the tip for two branches. Returning `feature3` in this case, cases an incorrect processing of the build. It used the wrong collection and also stored the build result in the wrong build result group.

```
show -s --pretty=%D HEAD
error: cannot run less: EDC5129I No such file or directory.
HEAD, origin/releaseGroup3, origin/feature3
```

PR #435, opened 2023-10-18
Improve calculation of branch name in detached head scenario

dennis-behm:
This is addressing the reported issue #432 

Enhancement contains:

* Message a warning if multiple references are found #143
* Flag the build as `error` if the branch name could not be obtained.

### DIFF
--- a/utilities/GitUtilities.groovy
+++ b/utilities/GitUtilities.groovy
@@ -73,20 +73,20 @@ def getCurrentGitDetachedBranch(String gitDir) {
 	}
 
 	String gitBranchString = gitBranch.toString()
-	def gitBranchArr = gitBranchString.split(',')
+	String[] gitBranchesArr = gitBranchString.split(',')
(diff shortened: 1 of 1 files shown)


utilities/GitUtilities.groovy
```diff
@@ -75,13 +75,29 @@ def getCurrentGitDetachedBranch(String gitDir) {
 	String gitBranchString = gitBranch.toString()
 	def gitBranchArr = gitBranchString.split(',')
```
M-DLB, line 76: For readability purpose, can you rename the `gitBranchArr` variable to something like `gitBranches`? Defining it as `String[]` would also help, but it's not required in groovy...

M-DLB (commented):
@dennis-behm a minor change for readability of the code, please. Thank you!

M-DLB (approved):
Looks all good - Thank you!

Issue #346, opened 2023-04-25
MFS. groovy


yamuna-tandra:
Hi, 

We have been using ibmdbb  code for one of our application. Where as in MaFS.groovy we are expecting TFORMAT pds to be copy into buildreport. Json which is not handling in existing code. 

Could you please help us to get the requirement. 

mfsPhase2.dd(new DDStatement().name("FORMAT").dsn(props.mfs_tformatPDS).options("shr").output(true).deployType(deployType))
	// mfsPhase2.dd(new DDStatement().name("DUMMY").dsn("${props.PROCLIB}(FMTCPY)").options("shr"))
	mfsPhase2.dd(new DDStatement().name("TASKLIB").dsn(props.SDFSRESL

dennis-behm:
Delivered via [zAppBuild Release 3.5.0](https://github.com/IBM/dbb-zappbuild/releases/tag/3.5.0)

Issue #36, opened 2020-08-25
MFS.groovy should use temporary dataset allocations for REFRD, REFOUT, REFIN


dennis-behm:
Today, there is a hardcoded dependency on a REFERAL dataset. Should use temporary files.

dennis-behm:
```
mfsPhase1.dd(new DDStatement().name("REFIN").dsn("&&REFERAL").options("${props.mfs_tempOptions} dir(10) lrecl(80) blksize(800) recfm(f,b)"))
mfsPhase1.dd(new DDStatement().name("REFOUT").dsn("&&TEMPPDS").options("${props.mfs_tempOptions} dir(5) lrecl(80) recfm(f,b)"))
mfsPhase1.dd(new DDStatement().name("REFRD").dsn("&&TEMPPDS").options("cyl space(5,5) unit(vio) old"))
```

bala-azha:
I have tested the temporary files for REFERAL as suggested in comments and it is working as expected.

jbyibm:
@dennis-behm 
Completely agree to open a PR because I REFERRAL an existing PDF from my IMS. Phase1 crashed in 3030 with a log size of zero. With this code ^^^ I was able to compile my MFS.

dennis-behm:
Delivered via [zAppBuild Release 3.5.0](https://github.com/IBM/dbb-zappbuild/releases/tag/3.5.0)

PR #436, opened 2023-10-18
Update MFS.groovy declaring the map as output

dennis-behm:
This is addressing #36, #346 and incorporates feedback from the field.

What is changed:
* The MFS source is defined as a build output in the DBB Build report. (Because the process generated environment specific formats, and the number for formats cannot be determined).
* The phase2 process becomes optional.

Commits need to be squash (started from `main` instead of `develop`)


languages/MFS.groovy
```diff
@@ -63,6 +73,8 @@ sortedList.each { buildFile ->
 			props.error = "true"
 			buildUtils.updateBuildResult(errorMsg:errorMsg,logs:["${member}.log":logFile])
 		}
+		
+		}
```
M-DLB, line 77: For better readability, can you change the indentation of the block between L65 and L75?
Thank you!

M-DLB (commented):
Looks good, just a small change for the indentation.

languages/MFS.groovy
dennis-behm: @M-DLB Applied pretty print.

M-DLB (approved):
Thank you!

Issue #395, opened 2023-08-08
SDSNEXIT required to prepare Db2 programs


FALLAI-Denis:
Hi,

To prepare a Db2 program (create the DBRM), you must use the DSNHDECP table which declares the codepage / ccsid of the target Db2 system.

The DSNHDECP table must be built and stored in the SDSNEXIT library when Db2 is installed.

It is therefore necessary to reference the SDSNEXIT library by means of the build of the Db2 programs, either in the Db2 preprocessor, or in the compilation of the program if the compiler has a Db2 coprocessor (as is the case for COBOL).

It would therefore be advisable to add an SDNSEXIT property in the datasets.properties file of zAppbuild.

SDSNEXIT should also be referenced by DDStatement TASKLIB in langage.groovy, before SDSNLOAD.

Thanks.

FALLAI-Denis:
Hi @dennis-behm ,

This seems ok to me.

PR #437, opened 2023-10-18
Introduce SDSNEXIT library for Assembler, Cobol and PLI language script

dennis-behm:
This is implementing issue #395 to accommodate different character translations / conversion for Db2 programs.

If **SDSNEXIT** it specified and the build file is a db2 program, the site specific SDSNEXIT library is added to the STEPLIB/TASKLIB to include the `dsnhdecp` module.


M-DLB (approved):
All seems OK to me. Thank you!

Issue #446, opened 2023-11-22
Missing exception handling in some of the groovy scripts


monvora:
When an exception occurs during build run e.g. a truncation error in a file (there is some row that text exceeds the column limit), the build is terminate unexpectedly. We need a way to handle this exception and set the build status properly.

dennis-behm:
This has been delivered via [zAppBuild Release 3.6.0](https://github.com/IBM/dbb-zappbuild/releases/tag/3.6.0)

Issue #441, opened 2023-10-24
Failures when uploading files to target libraries need to mark the build as failed


dennis-behm:
When files are uploaded to the target build libraries, zAppBuild is not evaluating the return code or checks for any exceptions. This can lead to the situation that the build status is clean and subsequent impact builds start of a more recent commit. 

https://github.com/IBM/dbb-zappbuild/blob/d177891a59a41b0947eacaf1e1e5a7ca199ac13a/utilities/BuildUtilities.groovy#L101-L108

dennis-behm:
Delivered via [zAppBuild Release 3.6.0](https://github.com/IBM/dbb-zappbuild/releases/tag/3.6.0)

PR #448, opened 2023-11-24
Handle exceptions thrown by the DBB runtime

dennis-behm:
This is implementing the #446 and #441 to improve the exception handling

* Build exceptions in the language scripts, will be centrally handled.
* When a build exception is caught, the build will finalise the build and skip any remaining elements on the build list.
* Transfer.groovy will not continue the build, but properly stop processing the build.


dennis-behm:
The latest [commit](https://github.com/IBM/dbb-zappbuild/pull/448/commits/bd6a348b3b81cd189164ac80e61cd50672c87123) also captures to handle any AssertionErrors that are thrown.

dennis-behm:
Added a new zAppBuild test case for the Mortgage application to make sure that the code always runs through the finalisation of build process when there is a BuildException, that includes truncation errors or other build exceptions.

@M-DLB ready for a final review.

M-DLB (approved):
Approved, thank you @dennis-behm

PR #449, opened 2023-11-24
Document the zAppBuild and DBB Toolkit versions in build log and build result

dennis-behm:
Capture the zAppBuild and DBB Toolkit version in the build log and build result, to understand the state of the build framework that was used to build the binaries.


M-DLB (approved):
Great enhancement, thank you!

Issue #452, opened 2023-11-29
Enhance processing of IMS programs


dennis-behm:
This request is about enhancing the support of processing IMS programs. Especially in mixed environments with CICS and IMS, this is not covered well.

So, let's collect requirements: 

**Calls to DLI interface**

Link phase: 

When running the program in CICS, it requires to pull the `CBLTDLI` from the `SDFHLOAD` library, but when running in an IMS environment (maybe IMS batch), there is no logic to add the SDFSRESL (IMS RESLIB) to pull the `DFSLI000` interface (CBLTDLI is just an alias) 

It requires a new flag to indicate that a program will be executed in an IMS context, such as an `isIMS` flag.

**Library concatenation when processing link cards in LinkEdit.groovy**

Link phase should be conditional to the "environment" flags as well. 

https://github.com/IBM/dbb-zappbuild/blob/d177891a59a41b0947eacaf1e1e5a7ca199ac13a/languages/LinkEdit.groovy#L113-L122

avalon2:
Hi Dennis, 
I would like to control full SYSLIB-Concatenation with the linkedit_linkEditSyslibConcatenation Property. Even better would be to have the possibility to use different Concatenations based on the Programm Language/Subsystem Combinations (e.g. CICS/DB2, PLI/DB2, PLI/IMS).  Although in our cases the SCEELKED (LE-Linklib) is always the first. It can happen that the DB2-Linklib or IMS-Linklib are alternating on 2nd Position in the Concatenation

a) PLI Programm with DB2
SCEELKED
DB2-Linklib
IMS-Linklib
Application Libs

b) PLI Programm with IMS
SCEELKED
IMS-Linklib
DB2-Linklib
Application Libs

avalon2:
Additionally, in the Compiler Step, we use a different SYSTEM-Parm if the Module contains DLI-Calls
a) IMS Program - SYSTEM(IMS)
b) Other Programs - SYSTEM(MVS)

dennis-behm:
> Additionally, in the Compiler Step, we use a different SYSTEM-Parm if the Module contains DLI-Calls a) IMS Program - SYSTEM(IMS) b) Other Programs - SYSTEM(MVS)

@avalon2 ,the capability exists today via file properties to set different SYSTEM parameters: 

```
# defining the defaults
pli_compileParms=PP(INCLUDE('ID(++INCLUDE)')),SYSTEM(MVS) 
# providing the override for IMS programs
[...]

PR #454, opened 2023-12-05
Add configuration options for IMS batch and online programs

dennis-behm:
This PR is addressing #452 .

For IMS batch or online programs, the DBB scanners don't detect any file attributes.

This PR is adding

* a new file property `isIMS` that is used to control the allocation of the DFSRESL library in the SYLSIB for the link phase in Assembler, Cobol, PLI  and LinkEdit language processors from where the linkage editor will pick up the appropriate language/runtime interfaces modules for Db2, DLI calls (primarily DFSLI000).

* The deployTypes settings are extended to define the deployType for programs with the isIMS flag turned on.

* Explanations of the file flags extended.

### DIFF
--- a/utilities/BuildUtilities.groovy
+++ b/utilities/BuildUtilities.groovy
@@ -411,6 +411,18 @@ def isMQ(LogicalFile logicalFile) {
 }
 
 /*
+ * isIMS - tests to see if the program is a DL/I program. If the logical file is false, then
+ * check to see if there is a file property.
+ */
+def isIMS(LogicalFile logicalFile) {
+	isIMS = false
+	String imsFlag = props.getFileProperty('isIMS', logicalFile.getFile())
+	if (imsFlag)
+		isIMS = imsFlag.toBoolean()
+	return isIMS
+}
+
+/*
  * getMqStubInstruction -
  *  returns include defintion for mq sub program for link edit
  */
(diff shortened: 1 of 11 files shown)


dennis-behm:
@M-DLB can you please check if you agree with the latest commit https://github.com/IBM/dbb-zappbuild/pull/454/commits/80e0753e3ebe86e42e826dded1c69874a6b57233 to add the capability to add pli compile parts depending on the new `isIMS` flag ?

See also - https://www.ibm.com/docs/en/epfz/6.1?topic=descriptions-system

utilities/BuildUtilities.groovy
```diff
@@ -558,6 +570,9 @@ def getDeployType(String langQualifier, String buildFile, LogicalFile logicalFil
 			} else if (isDLI(logicalFile)){
 				String dliDeployType = props.getFileProperty("${langQualifier}_deployTypeDLI", buildFile)
 				if (dliDeployType != null) deployType = dliDeployType
+			} if (isIMS(logicalFile)){
```
M-DLB, line 573: Is there an 'else' missing ?
And what happens for IMS transactions that do IMS DL/I, they are flagged as `deployTypeIMS`? Is it correct?
In the end, `deployTypeIMS` and `deployTypeDLI` are all pointing to `IMSLOAD`, so is there a need to distinguish `deployTypeIMS` and `deployTypeDLI`? I don't know the answer, just asking :)

M-DLB (commented):
One comment on the deploy type.

utilities/BuildUtilities.groovy
dennis-behm: @M-DLB Good catch, it is indeed cleaner with the `else if`. - Added with the last commit.

To your comment about the deployType for EXEC DLI - 

I intentionally changed the order for the deploy types, because they indicate the runtime environment to the deployment solution.

While EXEC DLI can be used for multiple runtimes, the `isCICS`, and `isIMS` flag are considered before the `isDLI` flag. See - https://www.ibm.com/docs/en/ims/15.4.0?topic=execution-translator-compiler-binder-options-required-exec-dli

It might need some refactoring to reevaluate the isDLI flag ...

M-DLB (approved):
Approved, thank you!

Issue #455, opened 2023-12-12
Reporting capabilities depend on metadata store connection / git repo in workspace


dennis-behm:
Today, the reporting capabilities are executed if centrally turned on via their properties.

The below section should validate their prerequisites :

* Reporting of external impacts is depending on the metadata store connection
* Reporting of concurrent changes is depending on the workspace containing a git repository / other references.

These two are generally not met in a user build scenario.

https://github.com/IBM/dbb-zappbuild/blob/d177891a59a41b0947eacaf1e1e5a7ca199ac13a/build.groovy#L643-L659

dennis-behm:
Delivered via zAppBuild Release 3.6.0 https://github.com/IBM/dbb-zappbuild/pull/456

PR #459, opened 2023-12-14
Validate prerequisites of reporting capabilities

dennis-behm:
This is implementing #455 and provides a more robust way when the build admin has centrally activated the reporting capabilities.


M-DLB (approved):
All is looking good!

PR #471, opened 2024-02-13
Validate System Datasets

dennis-behm:
This is a new feature helping during the setup phase of zAppBuild to validate that the defined system datasets in `datasets.properties` are correctly configured and present on the system where the build is performed.

### DIFF
--- a/build.groovy
+++ b/build.groovy
@@ -20,6 +20,7 @@ import groovy.cli.commons.*
 @Field def reportingUtils= loadScript(new File("utilities/ReportingUtilities.groovy"))
 @Field def filePropUtils= loadScript(new File("utilities/FilePropUtilities.groovy"))
 @Field def dependencyScannerUtils= loadScript(new File("utilities/DependencyScannerUtilities.groovy"))
+@Field def validationUtils= loadScript(new File("utilities/DatasetValidationUtilities.groovy"))
 @Field String hashPrefix = ':githash:'
 @Field String giturlPrefix = ':giturl:'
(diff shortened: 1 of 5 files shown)


build.groovy
```diff
@@ -268,14 +269,15 @@ options:
 	cli.r(longOpt:'reset', 'Deletes the dependency collections and build result group from the MetadataStore')
 	cli.v(longOpt:'verbose', 'Flag to turn on script trace')
 	cli.pv(longOpt:'preview', 'Supplemental flag indicating to run build in preview mode without processing the execute commands')
+	cli.cds(longOpt:'checkSystemDatasets', 'Optional flag to validate the presense of the defined system datasets. ')
```
M-DLB, line 272: Shouldn't it be `-csd` as the first letters of "Check System Datasets"?

M-DLB (commented):
One comment about the name of the option flag :)

build.groovy
dennis-behm: @M-DLB - `CSD` is already a known term from CICS. Should we rename the long option to `checkDatasets` instead ? 

I added the _system_ datasets, because it is really just the system datasets that can be validated in advance and not the additional build datasets that are created during build time.

M-DLB: Let's call it `checkDatasets `then, so `-cd` should be fine :)

FALLAI-Denis: Hi,
Or -ivp (Installation Verification Procedure)?

PR #473, opened 2024-02-14
Validate existence of workspace and outDir folder

dennis-behm:
Avoid failing builds by validating that the mandatory directories specified via `--workspace` and `--outDir` exist.


M-DLB (approved):
Thanks for the change, and approved!

Issue #474, opened 2024-02-14
Reporting capability creates empty files


dennis-behm:
Need to make sure, that file only are created if logical dependencies exists within the collection:

https://github.com/IBM/dbb-zappbuild/blob/dd59b05112791ef1ea5fc710d821cb214b90142d/utilities/ReportingUtilities.groovy#L156-L170

Current behaviour is creating report logs for all inspected DBB collections

![image](https://github.com/IBM/dbb-zappbuild/assets/28918869/aef57044-12f9-4a37-a336-948faeb98163)

dennis-behm:
Delivered via [zAppBuild 3.7.0](https://github.com/IBM/dbb-zappbuild/releases/tag/3.7.0)

PR #475, opened 2024-02-14
Generate reports only for impacted applications with impacted files

dennis-behm:
This is addressing the reported issue #474 to only generate reports for those impacted collection with identified impacted files.


M-DLB (approved):
Thank you @dennis-behm

PR #515, opened 2024-06-22
Update build.groovy

GeraldMit:
finalizeBuildProcess. optional second param error path issue

NPE avoidance for buildOrder (caught instead at required Properties)

### DIFF
--- a/build.groovy
+++ b/build.groovy
@@ -55,7 +55,7 @@ if (buildList.size() == 0)
 else {
 	if (!props.scanOnly && !props.scanLoadmodules) {
 		println("** Invoking build scripts according to build order: ${props.buildOrder}")
-		String[] buildOrderList = props.buildOrder.split(',')
+		String[] buildOrderList = props.buildOrder ? props.buildOrder.split(',') : []
 		String[] testOrderList;
 		if (props.runzTests && props.runzTests.toBoolean()) {
 			println("** Invoking test scripts according to test order: ${props.testOrder}")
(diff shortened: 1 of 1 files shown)


build.groovy
```diff
@@ -711,6 +711,9 @@ def createBuildList() {
 }
 
 def finalizeBuildProcess(Map args) {
+     finalizeBuildProcess(args, 0)
```
dennis-behm, line 714: I would suggest we follow the same strategy and access the attributes in the map instead of overriding the `finalizeBuildProcess` method

dennis-behm (changes requested):
@GeraldMit 

There are additional problems being introduced with these changes.

The finalise Method is accessing `count` from the args map: 

https://github.com/IBM/dbb-zappbuild/pull/515/files#diff-665cfd7c0dbe02e87a7dcb77c326a4e1a22b1fdd7e400977d6e6e49d77602a70R815

This should actually be tested if `count` is in the map.
```
	if(args.count) {println("** Total files processed : ${args.count}") }
```

In this section - we could then just pass the `startTime`, with the above update.
https://github.com/IBM/dbb-zappbuild/pull/515/files#diff-665cfd7c0dbe02e87a7dcb77c326a4e1a22b1fdd7e400977d6e6e49d77602a70R42

The `buildOrder` should actually be tested as a required build property: 
https://github.com/IBM/dbb-zappbuild/blob/a719842f8b9738133e375eaade8497289c30fc05/build-conf/build.properties#L115-L117 + https://github.com/IBM/dbb-zappbuild/blob/a719842f8b9738133e375eaade8497289c30fc05/build.groovy#L126-L127

So, I believe your problem has a different root cause.

dennis-behm:
This PR will be superseded by - https://github.com/IBM/dbb-zappbuild/pull/522

PR #522, opened 2024-06-28
Fix NPE on build process finalization in catch clause

dennis-behm:
Fix of the invocation of the finalizeBuildProcess in the catch block if an `AssertionError` is caught.


M-DLB (approved):
All good! Thank you @dennis-behm

PR #530, opened 2024-08-14
Uplift CRB.groovy to use UnixExec

M-DLB:
This PR adds the use of the new DBB API UnixExec, used to trigger a process on Unix System Services.
The UnixExec API also creates a Unix Record in the Build Report, which removes the need of a custom AnyType Record.


dennis-behm (approved):
@M-DLB LGTM.

I tested it with the supplied Mortgage Application, and it worked fine, BuildReport.json looks also fine with the Unix exec record.

Thanks for the update.

PR #538, opened 2024-09-04
Uplift of the zCEE3 language script

M-DLB:
This enhancement to the zCEE3.groovy script is to use the new DBB UnixExec API (version 2.0.2) which introduces the UnixRecord in the DBB Build Report.
Some changes were also made to properties, which should now be declared in the user's .profile (IBM_JAVA_OPTIONS environment variable).

### DIFF
--- a/build-conf/zCEE3.properties
+++ b/build-conf/zCEE3.properties
@@ -1,4 +1,9 @@
 # Releng properties used by language/zCEE3.groovy
+#
+# Important considerations when using Gradle with Java 11 on z/OS Unix System Services
+# IPv6 needs to be disabled by setting the IBM_JAVA_OPTIONS environment variable as follows:
+#    IBM_JAVA_OPTS="-Djava.net.preferIPv4Stack=true -Djava.net.preferIPv6Addresses=false -Dfile.encoding=UTF-8"
+# It is recommended to set this globally in the .profile file of the user that runs the build
 
 #
 # Comma separated list of required build properties for zCEE3.groovy
@@ -12,15 +17,6 @@ zcee3_shellEnvironment=bash
 # Absolute path to gradle executable on z/OS UNIX System Services
 # for instance: /var/gradle/gradle-7.6/bin/gradle
 zcee3_gradlePath=/var/gradle/gradle-8.10/bin/gradle
-
-#
-# JVM Options used with gradle
-# The use of IPv6 should be currently disabled for gradle and ava 11 on z/OS
-# This can be done by exporting the IBM_JAVA_OPTIONS environment variable as follows:
-# IBM_JAVA_OPTS="-Djava.net.preferIPv4Stack=true -Djava.net.preferIPv6Addresses=false"
-# Instead of specifying this information through a property, it is highly recommended
-# to set the value and export it from the .profile file of the user that runs the build
-# DEPRECATED - zcee3_gradle_IBM_JAVA_OPTIONS="-Djava.net.preferIPv4Stack=true -Djava.net.preferIPv6Addresses=false -Dfile.encoding=UTF-8"
 
 #
 # Flag to indicate if --debug option should be added to the gradle command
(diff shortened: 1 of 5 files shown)


build-conf/zCEE3.properties
```diff
@@ -11,10 +11,21 @@ zcee3_shellEnvironment=bash
...
 # Absolute path to gradle executable on z/OS UNIX System Services
 # for instance: /var/gradle/gradle-7.6/bin/gradle
-zcee3_gradlePath=
+zcee3_gradlePath=/var/gradle/gradle-8.10/bin/gradle
 
 #
-# JAVA Options used with gradle
-# for instance: zCEE_gradle_JAVA_OPTS=-Xmx512m -XX:MaxMetaspaceSize=256m -Dfile.encoding=UTF-8
-zcee3_gradle_JAVA_OPTS=-Dfile.encoding=UTF-8
+# JVM Options used with gradle
```
dennis-behm, line 17:
```suggestion
# JVM Options used with Gradle
```

build-conf/zCEE3.properties
```diff
@@ -11,10 +11,21 @@ zcee3_shellEnvironment=bash
...
+# The use of IPv6 should be currently disabled for gradle and ava 11 on z/OS
```
dennis-behm, line 18:
```suggestion
# Important considerations when using Gradle with Java 11 on z/OS Unix System Services
# IPv6 needs to be disabled by setting
#    IBM_JAVA_OPTS="-Djava.net.preferIPv4Stack=true -Djava.net.preferIPv6Addresses=false -Dfile.encoding=UTF-8"
# It is recommended to set this globally in the .profile file of the user that runs the build
# Alternatively you can use the deprecated build parameter
# zcee3_gradle_IBM_JAVA_OPTIONS="-Djava.net.preferIPv4Stack=true -Djava.net.preferIPv6Addresses=false -Dfile.encoding=UTF-8"
```

dennis-behm, line 18: I don't see `zcee3_gradle_IBM_JAVA_OPTIONS` being used in the zCEE3.groovy language script. Can you please explain?

languages/zCEE3.groovy
```diff
@@ -32,80 +32,67 @@ sortedList.each { buildFile ->
...
+                String errorMsg = "*! Error when searching for the 'api.war' file at location '${WarLocation}'"
                 println(errorMsg)
                 props.error = "true"
                 buildUtils.updateBuildResult(errorMsg:errorMsg)
             }   
         }
     }
 }
+
+def fileExists(String fileLoc){
```
dennis-behm, line 87: Do we really need a method for it? It is only used once and could be incorporated into the main logic.

dennis-behm (commented):
@M-DLB Can you please have a look to the below comments?

build-conf/zCEE3.properties
M-DLB: Yes it's deprecated. I should completely remove this reference.
The only way would be the .profile approach.

languages/zCEE3.groovy
M-DLB: It's also used by `CRB.groovy` I think, so it would make sense to put it in a utility script, so it can be used from several scripts.

M-DLB:
@dennis-behm Please take a look at the latest commit which contains some changes based on your feedback.

utilities/BuildUtilities.groovy
```diff
@@ -1021,4 +1021,23 @@ def isGeneratedzUnitTestCaseProgram(String buildFile) {
 	return false
 }
 
+/**
+ * Checks if a zFS file exists and if not, return false and set the Build re
```
dennis-behm, line 1025: @M-DLB This comment seems incomplete. Can you please have a look?

dennis-behm (approved):
LGTM

PR #551, opened 2024-10-14
Added documentation of object decks with deployType

M-DLB:
This PR adds a "deploy type" to all Object decks generated during the build process. Object decks are now documented in the DBB Build Report, with an object type "OBJ". Object decks can be then processed and potentially integrated in the archive.

### DIFF
--- a/languages/Assembler.groovy
+++ b/languages/Assembler.groovy
@@ -314,7 +314,7 @@ def createAssemblerCommand(String buildFile, LogicalFile logicalFile, String mem
 	assembler.dd(new DDStatement().name("SYSUT1").options(props.assembler_tempOptions))
 
 	// define object dataset allocation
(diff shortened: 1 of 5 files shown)


dennis-behm (changes requested):
I would prefer a new setting this name via a build property, similarly to the other deployTypes.

This is the sample for COBOL - 
https://github.com/IBM/dbb-zappbuild/blob/f3dfad68c6b727136830cd47c97a330b84fcc387/samples/application-conf/Cobol.properties#L76-L90

For COBOL, it could be `cobol_deployType_object`, or  `cobol_object_deployType`.

What are your thoughts on it?

M-DLB:
The problem is that we need to parse the build report afterward to extract these artifacts based on their deploy type. We could add a property for each type of artifacts, but it means we would need to add a property in packaging scripts to specify what is the deploy types associates to object decks. Do you want to add more properties?

dennis-behm (approved):
Thanks for making these adjustments to flag any OBJ as a potential build output, that could be part of the package.

Issue #548, opened 2024-09-30
REXX.groovy NullPointerException when some properties are not set


jbyibm:
Caught: java.lang.NullPointerException: Cannot invoke method toInteger() on null object
java.lang.NullPointerException: Cannot invoke method toInteger() on null object
	at REXX$_run_closure1.doCall(REXX.groovy:76)
	at REXX.run(REXX.groovy:28)

True for

- rexx_compileMaxRC
- rexx_linkEdit
- rexx_linkEditMaxRC

It could be good to have default values or make them mandatory in rexx_requiredBuildProperties

PR #552, opened 2024-10-14
Fix for REXX properties

M-DLB:
Fixing issue #548 
The PR adds some checks and use defaults values if these properties are not set. They are set by default in the REXX.properties file of application-conf folder.

### DIFF
--- a/languages/REXX.groovy
+++ b/languages/REXX.groovy
@@ -78,7 +78,6 @@ sortedList.each { buildFile ->
 		String needsLinking = props.getFileProperty('rexx_linkEdit', buildFile)
 		if (needsLinking && needsLinking.toBoolean()) {
 			rc = linkEdit.execute()
-			maxRC = props.getFileProperty('rexx_linkEditMaxRC', buildFile).toInteger()
 			if (props.getFileProperty('rexx_linkEditMaxRC', buildFile)) {
 				maxRC = props.getFileProperty('rexx_linkEditMaxRC', buildFile).toInteger()
 			} else {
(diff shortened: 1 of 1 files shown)


dennis-behm (changes requested):
Hi @M-DLB,

these updates feel odd to me. 

zAppBuild has the following capabilities:
* Enforce the presence of build properties via the `rexx_requiredBuildProperties` list in build-conf. Looks like that these properties are missing there.
* and the capability to set defaults - probably via the central `applicationDefaultPropFiles` setting in [build.properties](https://github.com/IBM/dbb-zappbuild/blob/f3dfad68c6b727136830cd47c97a330b84fcc387/build-conf/build.properties#L30-L54) if the customer is looking for centrally controlling it.

Any specific reasoning to hard-code the defaults in the script itself?

M-DLB:
You're right it's odd. It is just a workaround for people who are trying to use zAppBuild to build REXX programs, without having taken REXX.properties from `samples/application-conf/` and put that in the application's `application-conf`.
I'm happy to kill this PR, as long as we accept that we have the same problem with all our other scripts. For instance Cobol.properties in build-conf doesn't require `cobol_compileMaxRC`, `cobol_linkEditMaxRC` and `cobol_linkEdit` which are set in Cobol.properties of application-conf, and Cobol.groovy is using them.
What do you think?

dennis-behm:
@M-DLB 

> You're right it's odd. It is just a workaround for people who are trying to use zAppBuild to build REXX programs, without having taken REXX.properties from `samples/application-conf/` and put that in the application's `application-conf`.

Let's see this as a broader initiative to reconcile how properties are set. It is time to make use of `applicationDefaultPropFiles`  and reduce the settings in `application-conf`, and review the required build properties

dennis-behm:
@M-DLB This looks a bit messy. What happened with the rexx updates?

M-DLB:
I'll drop this PR, I made mistakes when reverting commits. Better drop this branch.

Issue #558, opened 2024-11-06
Aligning allocation options for work datasets for Cobol with recommendations


dennis-behm:
Cobol recommends the below allocation options for work datasets (SYSUTn):
https://www.ibm.com/docs/en/cobol-zos/6.4?topic=output-data-sets-used-by-compiler-under-zos

Todo - update Cobol.groovy settings

dennis-behm:
Delivered with the [zAppBuild Release 3.10.0](https://github.com/IBM/dbb-zappbuild/releases/tag/3.10.0)

PR #559, opened 2024-11-06
Update Cobol work datasets allocation options

dennis-behm:
This is implementing the recommended allocation options for COBOL work datasets. #558


M-DLB (approved):
Approved, thank you!

mhardin18 (commented):
Why are we hard coding this? Maybe it needs a separate property from cobol.tempOptions.

PR #565, opened 2024-11-08
REXX properties as required

M-DLB:
Fixing issue #548 
The PR adds some properties as required in the `rexx_requiredBuildProperties` property.


dennis-behm (approved):
Thanks - @M-DLB

PR #571, opened 2024-11-20
Adding logicalFileCollectionSizeLimit property to manage collection size during impact analysis.

mhardin18:
#
# Used in impact analysis to limit the size of the collections, and thus the resulting transactions at one time.
# The idea is to reduce load on the db2 database and on Heap space
#
logicalFileCollectionSizeLimit=500

### DIFF
--- a/build-conf/defaultzAppBuildConf.properties
+++ b/build-conf/defaultzAppBuildConf.properties
@@ -2,7 +2,7 @@
 # Default application properties used by zappbuild
 #
 #  treated as global defaults, if not overridden by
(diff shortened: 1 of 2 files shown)


dennis-behm:
@mhardin18 I understand the scenario you have been trying to circumvent based on the discussion you have opened. Are these updates still required?

mhardin18:
Well ultimately this didn't even solve my issue. So I would say they are not required.

dennis-behm:
Going ahead and close the PR

Issue #555, opened 2024-11-05
Enable Db2 binds for PL/I and Assembler


lucianap1111:
A script for binding these languages ensures that all dependencies and linkages are consistently managed in an automated fashion. This minimizes manual intervention, reducing errors and variability across builds.

A customized Groovy script would allow for handling each language's peculiarities, making the build process more robust and compatible.

The script can help modularize build processes, making it easier to maintain and update build configurations, particularly valuable as applications evolve or dependencies change.

dennis-behm:
Be aware that [Db2 bind process](https://github.com/IBM/dbb-zappbuild/pull/532) will be uplifted with the upcoming zAppBuild release.

dennis-behm:
Delivered via https://github.com/IBM/dbb-zappbuild/releases/tag/3.11.0

PR #583, opened 2024-12-09
Enable Db2 bind package and plan for Assembler and PLI

dennis-behm:
This is implementing the request #555 for Assembler and PLI. Includes minor cosmetically changes for Cobol.


M-DLB (approved):
Thank you @dennis-behm

PR #588, opened 2024-12-18
Detached Head Detection Improvement

### LINKED ISSUE #587: Update check for detached head state
Use `git symbolic-ref` instead of `git status` to detect a detached head state. See https://git-scm.com/docs/git-symbolic-ref

https://github.com/IBM/dbb-zappbuild/blob/db9fb093ac68ddd101146f7f2c4fd7494c7c2a6c/utilities/GitUtilities.groovy#L129-L146

kmcgreg5:
To avoid overhead associated with `git status`, changes git detached head detection to use the `git symbolic-ref` command. Head will be a reference to the branch if the repository is in a non-detached state, so the command will exit with a 0 RC. If the branch is in a detached state, HEAD will just contain the commit and not a reference. So the command will return a non-zero RC after failing to find a reference.

The `-q` option suppresses error output from a detached state.

This does change the return from false to true on an error scenario, in case that's important.

Resolves #587.


dennis-behm (approved):
Thank you @kmcgreg5 . Looks good to me. Thanks for implementing.

Issue #582, opened 2024-12-09
Implement impactBuildOnBuildPropertyChanges for individual artifact properties files


lauren-li:
Implement an option so that if only an individual artifact properties file has been modified since the last impact build (or baseline), the impact build will still be able to pick up the change and rebuild the relevant files.

Related PR: https://github.com/IBM/dbb-zappbuild/pull/110 (Ability perform impact builds on property changes.)

lauren-li:
This issue has been resolved by PR #605.

PR #605, opened 2025-04-22
Ability to perform impact analysis on changed/updated individual file properties

dennis-behm:
This expands the existing  capability to automatically add the corresponding file to the build list build properties change. 

This is specifically for the case when the individual file properties file (e.q. `epsmlist.cbl.properties`) in the application repository changes.

See #582


M-DLB (approved):
That looks good to me, thanks @dennis-behm

PR #607, opened 2025-04-22
Improve delete-only case in impactBuild scenario

dennis-behm:
These changes improve the scenario when a user only deletes a file from the repository. In this case, it tries to find a corresponding logical file in the DBB metadatastore before deleting it.


M-DLB (approved):
Approved, thank you @dennis-behm

PR #616, opened 2025-05-09
Documentation update for impact analysis on changed/updated individual file properties feature

dennis-behm:


build-conf/defaultzAppBuildConf.properties
```diff
@@ -68,8 +68,24 @@ generateDb2BindInfoRecord=false
...
+#  logical file tracked in the DBB Metadatastore. 
+#
+#  If an application manages 'individual file level properties' in 
+#  their application repository, this flag also controls
+#  impact analysis once individual file level properties are updated.
+#
+#  Property files must be stored within the application's git repository
+#
+# This feature can also be activated on application the application level
+#
```
lauren-li, line 88: @dennis-behm Thanks for drafting this up! I would like to propose a rewording/reformatting below, based on how existing property descriptions in this document seem to be formatted.

```
# impactBuildOnBuildPropertyChanges controls if changed build property files in the 'application-conf' folder (such as
# for COBOL, PL/I, and ASM programs) should cause an impact build.
# (See also: 'impactBuildOnBuildPropertyList'.)
#
# If set to true, zAppBuild performs an impact build upon changed build property files and adds additional dependencies
# to the logical file tracked in the DBB metadata store. 
#
# If an application uses individual file-level properties (i.e. individual artifact property files) in its repository,
# this flag also controls whether impact analysis is performed once individual artifact property files are updated.
#
# Property files must be stored within the application's Git repository.
#
# This feature can also be activated at the application level.
#   
# default = false

```

lauren-li (commented):
Thanks @dennis-behm for the updates! I've just added a couple more inline suggestions to refine the text with a max line length of 120 characters.

M-DLB (approved):

PR #627, opened 2025-05-22
Hotfix/long identify statement

monvora:
Fixed the issue reported that the identify statement parameters are too long if for example, the repository name is long.


dennis-behm (approved):
Thanks for the fixes, @monvora

PR #640, opened 2025-09-09
Use case when application teams add folders to applicationSrcDir

dennis-behm:
This is implementing a strategy on impactBuilds once the application team is adding a new folder to `applicationSrcDir` that was not in scope of the build previously (and therefore does not have a baseline reference that can be retrieved).

From now on, it will add all files to the build list as new files.


M-DLB (approved):
Approved.
Thank you @dennis-behm

Issue #274, opened 2022-11-04
Do not consider a moved file as a modified file


FALLAI-Denis:
Hi,

According to our analysis of the GitUtilities.groovy script, (not validated by a test), a moved-only file (Git action R100) is considered to be a file that is both **modified** and **renamed**, and it is added to both the `changedFiles` and `renamedFiles` lists.

We think that this should not be the case, because the construction (use case COBOL compilation), is in principle not linked to the location in the Git Repository, since the files are copied in PDS for then build them.

When a file is only moved (Git action R100) it should not be added to the `changedFiles` list, only added to the `renamedFiles` list, then task the build.groovy / _alangage_.groovy scripts to decide what to do on the files only renamed.

Files actually modified in addition to being moved (Git action R with a similarity index different from 100), must be added to the two lists changedFiles and renamedFiles.

FALLAI-Denis:
Hi, 

Problem stil present...

A "simply moved file" does'nt need to be built, (no change on source file, no change on binary outputs), but Collection needs to be updated to register is new path (and rescanned to find its dependencies...).

We have to modify:
* GitUtilities.groovy to generate a "newNameFiles" list : getChangedFiles and getCurrentChangedFiles functions
* ImpactUtilities.groovy to update Collection with "newNameFiles" list : UpdateCollection function

And all refernces to these functions...

PR #643, opened 2025-09-12
Support moving files in the repository without rebuilding them

dennis-behm:
This is implementing #274 to support customers that have onboarded to dbb-zappbuild , and want to move files in the repository to different folders after their migration.

When a change with a similarity index of 100 is detected, the DBB metadatastore is only updated, but the file is not added to the buildList.

### DIFF
--- a/utilities/GitUtilities.groovy
+++ b/utilities/GitUtilities.groovy
@@ -248,7 +248,7 @@ def getPreviousGitHash(String gitDir) {
 
 /*
  * getChangedFiles - assembles a git diff command to support the impactBuild for a given directory
- *  returns the changed, deleted and renamed files.
+ *  returns the changed, deleted, renamed and moved files.
  * 
  */
 def getChangedFiles(String gitDir, String baseHash, String currentHash) {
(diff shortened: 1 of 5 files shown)


utilities/GitUtilities.groovy
AnjaliAbraham, line 1: Comments on line 251, 261, 271 - returns the changed, deleted and renamed files. Shouldn't it have returns the changed, deleted, renamed  and moved files.

utilities/ImpactUtilities.groovy
```diff
@@ -491,12 +499,25 @@ def calculateChangedFiles(BuildResult lastBuildResult, boolean calculateConcurre
...
 
+		if (props.verbose) println "*** Moved files for directory $dir $msg:"
+		if (props.verbose) println "*** To be added to DBB Metadatastore, not rebuilt"
+		moved.each { file ->
+			if ( !buildUtils.matches(file, excludeMatchers)) {
+				(file, mode) = fixGitDiffPath(file, dir, false, mode)
+				movedFiles << file
+				if (props.verbose) println "**** $file"
+			} else {
+				if (props.verbose) println "**** $file is renamed, but is excluded from build scope. See excludeFileList configuration. No follow-up processing."
```
AnjaliAbraham, line 507: Line 510 - 				if (props.verbose) println "**** $file is moved, but is excluded from build scope. See excludeFileList configuration. No follow-up processing."

AnjaliAbraham (changes requested):
Hi Dennis - I have some comments. Please review

utilities/GitUtilities.groovy
dennis-behm: @AnjaliAbraham I have expanded the description for these methods. Well spotted.

utilities/ImpactUtilities.groovy
dennis-behm: I have updated the message that is written. Thanks.

AnjaliAbraham (approved):
Thanks Dennis for the changes

Issue #654, opened 2025-10-21
Validation of applying a file property override via Individual Property Files failing


dennis-behm:
There is an observed misbehaviour when processing individual property files.

Once the individual property files contain the same values, it starts to misbehave. The list of reported existing file property values does not match for the file in question, like below `epsmlist.cbl`. Further investigation required. 

```
*** Validating file property overrides
    *! An existing file property was detected for MortgageApplication/cobol/epsmlist.cbl
       Existing value cobol_compileParms=LIB
    *! Warning: MortgageApplication/cobol/epsmlist.cbl is already mapped as a file pattern as part of a file group wildcard possibly in file.properties.
    *! Warning: Please check the existing file property list below and fix it to contain the correct file properties.
    *! Existing File Property value and patterns :
    *!     cobol_compileParms = LIB for MortgageApplication/cobol/epsnbrvl.cbl
    *!     cobol_compileParms = LIB,SOURCE for MortgageApplication/cobol/epscmort.cbl
    *!     cobol_compileParms = LIB,SOURCE for MortgageApplication/cobol/epscsmrd.cbl
    *!     cobol_compileParms = LIB,SOURCE for MortgageApplication/cobol/epscsmrt.cbl
    *! Warning: Override for MortgageApplication/cobol/epsmlist.cbl could not be applied. Existing file property value will be used: cobol_compileParms = LIB
    *! Inserting warning message in BuildReport.json
```

PR #657, opened 2025-11-10
Address wrong behaviour on applying a file property override via Individual Property Files

dennis-behm:
This is addressing the defect reported in #654 . This is considered to be a work around, as it feels that the DBB toolkit is misbehaving when programmatically changing file properties.


AnjaliAbraham (commented):
Looks good, thanks Dennis

PR #658, opened 2025-11-10
Print verbose message if capturing static link dependencies is skipped

dennis-behm:
This adds extra tracing information in case a build is in error state, and customers wondering why the DBB metadatastore did not got updated.

### DIFF
--- a/utilities/ImpactUtilities.groovy
+++ b/utilities/ImpactUtilities.groovy
@@ -701,7 +701,9 @@ def saveStaticLinkDependencies(String buildFile, String loadPDS, String member, 
 		// Store logical file and indirect dependencies to the outputs collection
 		metadataStore.getCollection("${props.applicationOutputsCollectionName}").addLogicalFile( logicalFile );
 	} else {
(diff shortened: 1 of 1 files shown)


AnjaliAbraham:
Looks good

utilities/ImpactUtilities.groovy
```diff
@@ -700,6 +700,8 @@ def saveStaticLinkDependencies(String buildFile, String loadPDS, String member,
 
 		// Store logical file and indirect dependencies to the outputs collection
 		metadataStore.getCollection("${props.applicationOutputsCollectionName}").addLogicalFile( logicalFile );
+	} else {
+		if (props.verbose) println "*** Scanning load module for $buildFile was skipped due to prior condition (no metadatastore connection, build in error)."
```
M-DLB: Is this "no metadatastore connection, build in error" a complete list of errors, or suggestions? Maybe some more info should be added.

M-DLB (commented):
Just one comment

M-DLB (changes requested):

Issue #663, opened 2025-12-03
assembler_cicsprecompilerParms not referenced in Assembler.groovy


lauren-li:
The parameter `assembler_cicsprecompilerParms` is defined in [application-conf/Assembler.properties](https://github.com/IBM/dbb-zappbuild/blob/main/samples/application-conf/Assembler.properties), but is not referenced in [Assembler.groovy](https://github.com/IBM/dbb-zappbuild/blob/main/languages/Assembler.groovy). Instead, `Assembler.groovy` uses a misspelling of the parameter, `assember_cicsprecompilerParms` (note the missing "l" in "assembler): https://github.com/IBM/dbb-zappbuild/blob/main/languages/Assembler.groovy#L283

Note: The preceding parameter in the `Assembler.groovy` script, `assember_cicsprecompiler`, also contains a misspelling of `assembler`, but it is consistent with how the parameter is defined in [build-conf/Assembler.properties](https://github.com/IBM/dbb-zappbuild/blob/main/build-conf/Assembler.properties#L57).

PR #664, opened 2025-12-03
Address incorrect build variable reference for CICS precompile in Assembler

dennis-behm:
This is addressing #663


lauren-li (approved):
Looks good to me - Thanks @dennis-behm for addressing issue #663 so quickly!

Issue #674, opened 2026-02-17
Exception in case no user supplied baseline ref is returned by the utility method


dennis-behm:
There is a problem when having multiple directories on the applicationSrcDirs property in combination with the baselineRef option.

When no user-supplied baseline ref is provided for a applicationSrcDirs, fails with a NullPointerException.

https://github.com/IBM/dbb-zappbuild/blob/4ba55c9cdbd9103ea83f85c36c18dd49e476cf26/build.groovy#L767

PR #675, opened 2026-02-20
Address NullPointerException if the users has not specified a baseline hash for directory

dennis-behm:
This addresses defect #674  when multiple directories are on applicationSrcDirs, but the user does not have specified a baseline ash for each directory, like outlined in [docs/BUILD.md](https://github.com/IBM/dbb-zappbuild/blob/main/docs/BUILD.md#perform-impact-build-by-providing-baseline-reference-for-the-analysis-of-changed-files)


wfezzani:
I suggest the following change which has been tested and validated:
```
String baselineRef = props.baselineRef ? buildUtils.getUserProvidedBaselineRef(dir) : null
String userBaselineRef = baselineRef ? userBaselineRef.replaceAll("origin/","")
```

PR #676, opened 2026-05-21
Null Check on impacted files without a filename

dennis-behm:
If a user has scanned files that don't have a proper basename, the logicalFile may not have a fileName. The impact analysis may exit out with a NPE.

### DIFF
--- a/utilities/ImpactUtilities.groovy
+++ b/utilities/ImpactUtilities.groovy
@@ -90,31 +90,38 @@ def createImpactBuildList() {
 
 				impacts.each { impact ->
 					def impactFile = impact.getFile()
-					if (props.verbose) println "** Found impacted file $impactFile"
-					// only add impacted files that have a build script mapped to it
-					if (ScriptMappings.getScriptName(impactFile)) {
-						// only add impacted files, that are in scope of the build.
-						if (!buildUtils.matches(impactFile, excludeMatchers)){
+					if (impactFile != null) {
+						if (props.verbose) println "** Found impacted file $impactFile"
+						// only add impacted files that have a build script mapped to it
+						if (ScriptMappings.getScriptName(impactFile)) {
+							// only add impacted files, that are in scope of the build.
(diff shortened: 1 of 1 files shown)


utilities/ImpactUtilities.groovy
```diff
@@ -90,31 +90,38 @@ def createImpactBuildList() {
...
+						} else {
+							String warningMsg = "*! $impactFile is impacted by changed file $changedFile, but is not added to build list, because it is not mapped to a language script."
+							buildUtils.updateBuildResult(warningMsg:warningMsg)
+							println(warningMsg)
 						}
-					} else {
-						String warningMsg = "*! $impactFile is impacted by changed file $changedFile, but is not added to build list, because it is not mapped to a language script."
+					}
+					else {
+						String warningMsg = "*! The impacted file does not have a file name. RAW impact file object $impact."
```
M-DLB: Do you really want to print the logical file in its "raw" format? Maybe you can display its properties when they are not null, in a pretty way.

M-DLB (commented):
One comment

utilities/ImpactUtilities.groovy
dennis-behm: @M-DLB please checkout the updates in  https://github.com/IBM/dbb-zappbuild/pull/676/commits/84bb225332f2edf5e58210fb05d03c8f1f38a28a

utilities/ImpactUtilities.groovy
```diff
@@ -121,7 +121,7 @@ def createImpactBuildList() {
 						}
 					}
 					else {
-						String warningMsg = "*! The impacted file does not have a file name. RAW impact file object $impact."
+						String warningMsg = "*! The impacted file does not have a file name. impact file in JSON representation: ${impact.toJSON()}."
```
M-DLB, line 124: Would it be more readable if you print all the important fields by yourself?

M-DLB (approved):

PR #678, opened 2026-07-02
Update .gitattributes removing git-encoding

ijmitch:
working-tree-encoding is the standard option for specifying the encoding - obviously intentionally overridden by zos-working-tree-encoding when using the Git for z/OS port.

`git-encoding` has been obsolete since Git 2.26. Defaulting everywhere but z/OS to utf-8 should obviously be fine.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,22 +2,22 @@
 * text eol=lf
 
 # file encodings
-*.cpy zos-working-tree-encoding=ibm-1047 git-encoding=utf-8
-*.cbl zos-working-tree-encoding=ibm-1047 git-encoding=utf-8
-*.bms zos-working-tree-encoding=ibm-1047 git-encoding=utf-8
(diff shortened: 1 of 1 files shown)


ijmitch:
To answer my own question above, https://git-scm.com/docs/gitattributes actually says:

>Use the working-tree-encoding attribute only if you cannot store a file in UTF-8 encoding and if you want Git to be able to process the content as text.

which tells me you should not need a working-tree-encoding for utf-8.

lauren-li:
I don't have write access to this repo to do approvals, but just wanted to comment that I tested the `.gitattributes` file on my fork of this repo with and without `working-tree-encoding=utf-8` (making and viewing changes to some COBOL, copybook, BMS, `.properties`, and `.yaml` files as a sample), and the transfers between the remote Git repo, z/OS, and my local workstation seem to all work as before.

M-DLB (approved):

